PR #29, opened 2020-02-06
First CSV examples (structural)

mccalluc:
**On hold: Come up with field description first:** https://docs.google.com/spreadsheets/d/1-PB76iHqKF6f-MCKwVXdpP-k70u3vOI1RWlgT2TMFQU/edit#gid=0

---

The CSVs here are exported from a [google sheet](https://docs.google.com/spreadsheets/d/169pTK91ZydbUZVJYTIvs3jX6sotllIddYB9MY55AJNc/edit#gid=754910469),
but once this PR has buy-in, this repo will be the single source of truth, and that spreadsheet will be deleted, to avoid confusion.
(But while we're reviewing this PR, feel free to edit and re-export from there.)

With these samples, I am only trying to pin down the high-level structure of the CSVs:
What will the column names be? What IDs need to be supplied? How are entities across the provenance chain connected?

The actual metadata columns will be coming soon, but we don't need to block on that.

### DIFF
--- a/test/fixtures/csv/basic/README.md
+++ b/test/fixtures/csv/basic/README.md
@@ -1,0 +1,8 @@
+The CSVs here are exported from a [google sheet](https://docs.google.com/spreadsheets/d/169pTK91ZydbUZVJYTIvs3jX6sotllIddYB9MY55AJNc/edit#gid=754910469),
+but once this PR has buy-in, this repo will be the single source of truth, and that spreadsheet will be deleted, to avoid confusion.
+(But while we're reviewing this PR, feel free to edit and re-export from there.)
+
+With these samples, I am only trying to pin down the high-level structure of the CSVs:
+What will the column names be? What IDs need to be supplied? How are entities across the provenance chain connected?
+
+The actual metadata columns will be coming soon, but we don't need to block on that.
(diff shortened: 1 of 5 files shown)


mccalluc:
From Nils:
- hubmap ID is generated during submission: it can't be in the form... unless it changes. Nils will bring up in call.
- only record one parent, not the full hierarchy (top sample will duplicate this information: root=parent)
- assayfile: sample cols, assay fields, file it corresponds to.
- record the donor at every level, as a sanity check.

mccalluc:
@jswelling : Adding you on this. I think Nils would prefer for the ID to be known and stated in this CSV, but that's not possible right now.

jswelling:
We know that each data submission can include multiple datasets, with different assay modalities or even multiple instances with the same modality.  In a CSV representation, maybe we have 1 dataset per line?

CSVs are bad at representing lists.  Key-value tables, like an INI file or similar, might work around this.  Lists will be important because a modality may include multiple pieces of equipment and/or software.

If we have a 'metadata manifest' as proposed, and there is a CSV file indicating which datasets are to be derived from a given data submission, how are elements of the metadata manifest assigned to datasets?  Presumably a metadata element can be relevant to multiple datasets in the same data submission.  So far different datasets can be distinguished by directory tree within the data submission but no promises have been made that that will continue to be true.

mccalluc:
@jswelling : Thanks! Some follow-up questions: Feel free to shift to Slack if that's easier, or it can stay here.

> We know that each data submission can include multiple datasets

Not sure about this, but it might just be definitions: I am thinking that one data submission equals one "data set" equals one Neo4j node: There might be many files under this, or there might be derived data sets after processing.

> CSVs are bad at representing lists

Agreed... but we don't think we can get the labs to submit data in a more appropriate format. In lieu of doing it the right way:
- Rows in the CSV may be very denormalized, with lots of repeated values.
- If metadata doesn't need to be indexed, it doesn't need to be here. Lots of instrument specific data might be in standalone files which we won't worry about for now.
- In the worst case, we can have delimited value in the cell. (But I will fight against this!)

> and there is a CSV file indicating which datasets are to be derived from a given data submission

Same misunderstanding on my side: I'm thinking datasets and submissions are 1-1?

mccalluc:
This work was moved to ingest-validation-tools

PR #30, opened 2020-02-06
Add JSON fixtures

mccalluc:
@jswelling : This is intended to complement #29: On that PR, which concerns the external format, I'd like to get Nil's feedback first. On this PR, which is about the internal formats, I'd like to get your feedback first, and then I'll get his. ... they need to be in sync eventually.


mccalluc:
Closing. This was taken over by ingest-validation-tools.

PR #54, opened 2020-05-03
Single metadatatsv data collection 202005

jswelling:
Added a simple data collection type for a single -metadata.tsv alone, for developer use


jswelling:
But this was supposed to be against devel, not master.

PR #78, opened 2020-06-11
Welling symlinks filepaths 202006

jswelling:


jswelling:
@pdblood you're right, actually- I think I can fit the install of the local cwltool into requirements.txt .  It supports a wider syntax than I understood.  Will do so before I merge the branch.

jswelling:
Ok @pdblood, I've confirmed that there is no good way to include our cwltool in requirements.txt .  The problem is that "pip install -r path/to/requirements.txt" has a very poor understanding of relative paths in requirements.txt.  It's worth discussing though.

PR #80, opened 2020-06-16
Use pipeline output manifests, if present

mruffalo:
This PR implements filtering of the data files produced by analysis pipelines, leaving full results on disk but only including selected files in indexes and downstream steps (e.g. the portal UI).

### DIFF
--- a/src/ingest-pipeline/airflow/dags/codex_cytokit.py
+++ b/src/ingest-pipeline/airflow/dags/codex_cytokit.py
@@ -15,7 +15,11 @@ from airflow.hooks.http_hook import HttpHook
 
 import utils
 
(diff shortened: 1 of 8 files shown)


mruffalo:
Patterns are written in order -- the [schema](https://github.com/hubmapconsortium/ingest-pipeline/pull/80/commits/addfa5c4dab0615400bdba41a242b0b6ba9a8a05) specifies an array of objects. Metadata is returned by the first matching pattern, by design, similar to lists of URL patterns in packages like Django. It's up to the person writing the manifest to make sure that the order of patterns is appropriate if multiple patterns could match a certain file path.

jswelling:
FYI: there is a test at start-up time for the ingest-pipeline plug-in that verifies the presence of needed configuration parameters.

PR #82, opened 2020-06-17
Hotfix: read correct ontology term from manifest

mruffalo:
Found this while checking against schemas; this fix was only in a Git stash which happened to be what I tested against.


jswelling:
Looks good to me.

PR #96, opened 2020-06-26
Add scATAC-seq DAG

mruffalo:
I'm sure there are a lot of omissions in this -- I don't have a good handle on everything required to add a new data type along with this DAG. This seems ready to make a PR and ask for feedback though.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/sc_atac_seq_snare.py
+++ b/src/ingest-pipeline/airflow/dags/sc_atac_seq_snare.py
@@ -1,0 +1,340 @@
+import os
+import json
+import shlex
+from pathlib import Path
+from pprint import pprint
+from datetime import datetime, timedelta
+
+from airflow import DAG
+from airflow.operators.bash_operator import BashOperator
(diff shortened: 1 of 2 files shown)


jswelling:
Looks like I was wrong about the necessity to explicitly create the cwl_out subdir, since I think the CWL workflow knows to create its --output_dir.  That means the only needed change would be to the collection_type string in workflow_map.yml .

jswelling:
Oh, one other comment- the dataset_types set in t_send_create_dataset needs to be one unique string, and different from the assay type defined in the metadata.tsv file.  In the portal, the original dataset will have a dataset type defined by ingest-api and I think always matching the assay name from the metadata.tsv file.  The dataset type for the derived dataset the workflow is producing should suggest that assay type and what operations were performed to produce the derived data.

mruffalo:
Thanks for the feedback -- I'm not sure I understand everything about the `collection_type` key in `workflow_map.yml`. I'm guessing that changing this to `'single_metadatatsv'` is *not* appropriate -- so I need to add another `DataCollection` subclass?

PR #100, opened 2020-07-07
Support for setting hubmap-commons version via COMMONS_BRANCH env

jswelling:


yuanzhou:
Tested with localhost build with all other APIs and gateway. Worked well! And it turns out that in addition to a commons branch, we can also specify a tag or a commit for the build.

PR #141, opened 2020-08-18
Additional workflow function cleanups

mruffalo:
This PR addresses some (relatively minor) annoyances that I encountered while adding new DAGs -- some of which are my own doing.

I previously added some convenience functions for finding absolute `Path`s to workflows on disk, and I was exactly wrong about the nicest way to implement these, in terms of whether each function should accept an `Iterable[Path]` or a varargs `*paths`. Fix those after some more experience with how these functions are actually used.

Also add a single function that constructs a `shlex.quote`'d version of a command string, which handles conversion to `str` in one place. This is much nicer in terms of not having to worry about types of things in command string lists, especially with `Path` objects. Then, use that function to clean up conversion to `str` in all command list definitions.


mruffalo:
I tend to prefer `path_variable / "foo/bar"` rather than using the `/` operator multiple times between constants -- the latter reminds me of `string_variable + "constant string 1" + " " + "constant string 2"`. I'm not sure there's any formal style that's recommended, though; `pathlib` isn't even mentioned in PEP 8.

I really like using `pathlib` because object-oriented paths can make code cleaner and easier to read (in my opinion), and I feel like `path_variable / "foo" / "bar" / "baz"` does the opposite.

PR #144, opened 2020-08-24
Bump ome-tiff-pyramid submodule to v1.3b1

mruffalo:
This uses a development build of `bioformats2raw` as @ilan-gold described in https://github.com/hubmapconsortium/portal-ui/issues/1028#issuecomment-676679335, and this is ready for testing.

I ran `/opt/bioformats2raw/bin/bioformats2raw` in the new `hubmap/ome-tiff-pyramid:1.3b1` Docker image and verified that `bioformats2raw` now accepts a `--downsample-type` option -- this was not listed in the version included in `hubmap/ome-tiff-pyramid:1.2`. I didn't test beyond that, though.


ilan-gold:
@mruffalo I think one of us is missing something.  Looking at the beta branch changes here, it looks like the `--downsample=CUBIC` parameter is used all the time but we only want it used for IMS data.  Is is possible to make it configurable via CWL or the like?

PR #151, opened 2020-08-28
Welling multi parent 202008

jswelling:
Adds a DAG called 'launch_multi_analysis' which is intended to be manually triggered with a manually-entered JSON string containing run information.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/launch_multi_analysis.py
+++ b/src/ingest-pipeline/airflow/dags/launch_multi_analysis.py
@@ -2,6 +2,7 @@ import sys
 import os
 import json
 import shlex
+import ast
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


jswelling:
@mruffalo and @pdblood I went ahead and merged this because I need to move on to the next bit, but thought I'd show you what's going on.  There is no way to trigger multi-dataset runs from ingest-ui so this DAG is intended to be triggered manually from the Airflow UI.  Up-to-date versions of the Airflow UI present a field for JSON info to be fed as dag_run.conf to the triggered dag_run.  The user will put a JSON string like:

    {"uuid_list":["HBM867.FZZD.797","19f22c9a17543b08405063efb59781f1"],"collection_type":"bulkatacseq_collection"}

into that field.

PR #155, opened 2020-09-11
codex_cytokit DAG: fix build_cwltool_cmd2

mruffalo:
I introduced a bug in 2d1e5b4ead70ed8f28689d53a8eae91610c132ad, due to bit rot in rebasing that commit too many times.

Fix trying to use a `str` path like a `Path` object.


jswelling (approved):
Doesn't that evaluate to the same thing?  I agree that the new version is better, but is there evidence that this actually broke anything?

mruffalo:
`os.path.join` returns a `str`, which blows up when we try to use `data_dir` in a way that requires a Path object in the `command` list.

(The last element of `command` is `data_dir / 'output/extract/expressions/ome-tiff'`.)

jswelling:
Ah, right- I wasn't thinking about how the downstream logic was set up.  Merging.

PR #157, opened 2020-09-11
Version zhou

yuanzhou:
Individual build script to handle the localhost build/start/stop. 
Gateway calls this script when building with all other APIs.

@jswelling I've merged the `VERSION-zhou` branch into `devel` across all other repos. And I'll do a DEV build. Please let me know if you have any questions.


yuanzhou:
FYI @jswelling I redeployed on DEV with testing the `VERSION-zhou` changes across all other repos and it worked fine.

jswelling (changes requested):
I think the expectation is that absent_or_newer() is defined because this script is called from a script that defines it, but I think that's bad practice, especially since the script that calls it lives in an entirely different repo.  I think that either absent_or_newer() should get defined in this script, or its presence should be tested for.  The latter could be accomplished by including the following just before the call to absent_or_newer:
```
( LC_ALL=C type absent_or_newer | grep 'function' ) || echo 'absent_or_newer is not defined'
```

yuanzhou:
@jswelling thanks for reviewing. Apparently I forgot to add the absent_or_newer() definition in this script. I've fixed the bug.

PR #165, opened 2020-09-30
Welling localhost updates 202009

jswelling:
This set of updates changes localhost/docker mode to use a Docker container built 'from scratch' using python:3.7-slim-buster rather than the now-unsupported puckel/docker-airflow container.  This method also uses a version of Airflow installed with pip, which is more like the situation on the PROD &etc. environments.  The downside is that more python code is installed when the container is started up, so the time before the Airflow UI becomes live is increased over the older version.

This version is integrated with the recent updates to the gateway/hubmap-docker.sh mechanism.


pdblood (approved):

yuanzhou:
@jswelling I did a localhost build and everything worked fine with the image building and the ingest-pipeline container started with no problem too. Only two tweaks that I found:

1. The `docker-compose.stage.yml` is missing since I found other yaml files for localhost/dev/test/prod.
2. The `restart: always` setting should be removed from the `docker-compose.yml` base yaml file. The dev build doesn't need this, and all other environments will overwrite it in their own yaml file.

jswelling:
Agreed on both counts @yuanzhou - I'll make the mods.

jswelling:
@yuanzhou requested mods have been added.

yuanzhou (approved):

PR #166, opened 2020-09-30
Add a schema to launch_multi_analysis DAG

jswelling:
Multi-dataset analyses are currently started manually, by triggering the launch_multi_analysis DAG with a JSON string specifying the input datasets and the analysis.  This PR adds schema testing to the dag so that errors in that JSON string can produce sensible messages.


pdblood (approved):
Approved

PR #176, opened 2020-10-27
Welling/remove thirdparty tifffile

jswelling:
When this started we couldn't get a good version of 'tifffile' from Pypy, but it is now available.  That was the only code in the 'thirdparty' subdirectory, so I have deleted that subdirectory and replaced all references with a version loaded via requirements.txt .


build_number
```diff
@@ -1 +1 @@
-477
+591
```
icaoberg, line 1: @jswelling do we increase the build number manually iff we change the requirements file? or do we have to increase it every time?

icaoberg (approved):

PR #219, opened 2021-01-29
SPRM-to-AnnData Update for HuBMAP UI

ilan-gold:
This replaces the current SPRM pipeline of sprm-to-json with https://github.com/hubmapconsortium/portal-containers/pull/58.  Now all outputs will be AnnData and any future SPRM outputs should be run against this container.   Should I remove the updates to the other two repos or are those OK?  Is there anything else that needs to be done for getting this into this repo?

### DIFF
--- a/src/ingest-pipeline/airflow/dags/codex_cytokit.py
+++ b/src/ingest-pipeline/airflow/dags/codex_cytokit.py
@@ -55,6 +55,7 @@ with DAG('codex_cytokit',
         Path(pipeline_name, 'pipeline.cwl'),
         Path('portal-containers', 'ome-tiff-offsets.cwl'),
         Path('portal-containers', 'sprm-to-anndata.cwl'),
+        Path('portal-containers', 'sprm-to-json.cwl'),
     )
 
     def build_dataset_name(**kwargs):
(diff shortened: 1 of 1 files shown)


ilan-gold:
@jswelling I added back in the `sprm-to-json` for continuity even though we won't be using it anymore.

jswelling (commented):
Are we going to run both the sprm-to-anndata.cwl and the sprm-to-json.cwl workflows?  I think the change to codex_cytokit.py may require some mods beyond simply updating the cwl_workflows list.

ilan-gold:
@jswelling Yes that would be the idea.  Wasn't sure about what the mods would be, but feel free to check out the branch/give guidance to me.

jswelling (changes requested):
I've confirmed that the addition of the new CWL workflow will require the addition of a corresponding set of blocks to the DAG file codex_cytokit.py .  This is necessary because we haven't been able to find an alternative to manually unrolling the workflow steps, but it looks really easy in this case.  Please take a look and see if you can attempt the edit.  I'm happy to make corrections after the merge if necessary, during testing on DEV.

ilan-gold:
@jswelling I think I did it?  Could you take a look and if you feel comfortable, feel free to merge.

PR #221, opened 2021-02-02
Jpuerto psc/globus auth

jpuerto-psc:
What I Did:
Added functionality to lock Airflow behind an authentication routine. The authentication routine mirrors the routine that is used for other HubMap resources. That is, Oauth2 authentication with Globus.

Why I Did It:
This was done to allow external users to access Airflow resources. This also allows the resources to use the same user list as the HubMap resources.

How to Test:
Go to https://hivevm191.bridges.psc.edu:5555
From here, you should be redirected automatically to the Globus log-in page to authenticate with your institution's information.
After entering your information, you should be redirected back to the Airflow dashboard.

Patch Notes:
This change adds a few pip packages to the requirements.txt file.
This change also adds some configuration variables to the airflow.cfg file that MUST be used to be able to run appropriately.


jswelling (changes requested):
It looks like a couple of submodule updates got included in the PR, probably because I didn't clear them out correctly when I moved the previous branch off DEV. Is there a convenient way to remove them from the commit?
Also, I'm not sure the empty __init__.py is necessary under Python3.

jswelling (approved):
Perfect, Thanks!

PR #225, opened 2021-02-08
Update scATAC-seq pipeline, consolidate DAGs

mruffalo:
This PR adds a DAG generator for scATAC-seq, similar to the DAG generator for scRNA-seq added in #222. This uses a unified scATAC-seq pipeline and removes the need for multiple submodules pointing at different commits.


jswelling (approved):
Looks very efficient!  Do we need to start thinking about splitting utils.py?  It's getting pretty big.

PR #242, opened 2021-03-02
CODEX pipeline v2.0b1 (incl. stitching), separate SPRM and viz symlink archiving

mruffalo:
This updates the CODEX pipeline to a beta release `v2.0b1`, which includes stitching of tiles into a single image.

Additionally, SPRM is now its own pipeline with its own CWL workflow, as is the extra step of archiving symbolic links for the HuBMAP UI. This involved some significant additions and modifications to the `codex_cytokit.py` DAG, including referring to workflows by name and not number, to avoid any possibility of errors in renumbering workflows by hand.

A final `v2.0` release of the CODEX pipeline will follow shortly, either handling images that are already stitched, or after confirming that this will not be necessary. This combination of versions (`codex-pipeline`, `sprm`, `create-vis-symlink-archive`) seems ready for testing though.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,9 @@
 [submodule "src/ingest-pipeline/submodules/ingest-validation-tests"]
 	path = src/ingest-pipeline/submodules/ingest-validation-tests
 	url = git@github.com:hubmapconsortium/ingest-validation-tests.git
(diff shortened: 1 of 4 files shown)


jswelling (approved):
OK, let's merge it and test on DEV.

Issue #211, opened 2021-01-29
Data migration into AWS


khanshawPSC:
https://app.asana.com/0/0/1199513842012759/f

Milestone: Demonstrate pipeline execution on AWS and leveraging STRIDES.
STRIDES: Identify metrics for meaningful comparison of AWS and PSC on-prem
STRIDES: Identify a set of pipelines for execution
STRIDES: Execute containerized implementations of the pipelines on AWS and PSC on-prem resources
Demonstrate execution of assay-specific pipelines on AWS, launched from the HuBMAP Portal.
Document per-job accounting data from the runs on AWS and STRIDES.
Completion criteria: Document execution of at least 3 pipelines running on AWS and their associated costs through STRIDES.

jpuerto-psc:
Able to demonstrate data transfer from PSC Hive resources to AWS resources using the Globus Transfer API (and user based authentication).

The next step is to implement this within the context of an Airflow workflow.

jpuerto-psc:
https://github.com/hubmapconsortium/ingest-pipeline/tree/jpuerto/globus-transfer

Branch created to track this implementation.

khanshawPSC:
Possibly test example by the end of this sprint.

jpuerto-psc:
Initial DAG implementation complete.

Work will continue on adding configurability to the DAG so that users can set which files/directories they would like to transfer.

jpuerto-psc:
@khanshawPSC This can be closed.

PR #270, opened 2021-04-02
Jpuerto/globus transfer

jpuerto-psc:
- What I Changed
-   Adds a form to the HubMap API Plugin
-   Adds a DAG that takes a list of transfer items and performs a Globus transfer from Hive to AWS
- Why I Changed It
-   As per Issue #211 - this PR adds dynamic globus transfers from within a DAG context
- Revision
-   Minor

Additional Notes:
The structure of the configuration is as follows:
`{
"transfer_items": [
  {
    "src": "source directory or file path"
    "dest": "destination directory or file path"
    "recursive": "whether the transfer should be done recursively (true/false)"
  }
]
}`


jswelling (approved):
Wow, that's actually quite compact!

Issue #315, opened 2021-05-11
The assay name service should be used to canonicalize names before looking them up in workflow_map.yaml


jswelling:

jswelling:
Should be resolved by https://github.com/hubmapconsortium/ingest-pipeline/pull/316 .

jswelling:
Resolved by PR #316

PR #316, opened 2021-05-12
Try to canonicalize assay type using TypeClient before lookup in work…

jswelling:
…flow_map

Historically the assay type found in the metadata has been used 'raw' during lookup in workflow_map.yaml to determine which workflow(s) to use for downstream processing.  This PR adds canonicalization via hubmap_commons.type_client.TypeClient .

### DIFF
--- a/src/ingest-pipeline/airflow/dags/utils.py
+++ b/src/ingest-pipeline/airflow/dags/utils.py
@@ -20,6 +20,8 @@ from airflow.hooks.http_hook import HttpHook
 from cryptography.fernet import Fernet
 
 from hubmap_commons.schema_tools import assert_json_matches_schema, set_schema_base_path
+from hubmap_commons.type_client import TypeClient
+
 
 import cwltool  # used to find its path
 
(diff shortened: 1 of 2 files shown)


jswelling:
ref https://github.com/hubmapconsortium/ingest-pipeline/issues/315

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -1050,13 +1053,45 @@ def _get_workflow_map() -> List[Tuple[Pattern, Pattern, str]]:
...
 
 
+def _get_type_client() -> TypeClient:
+    """
+    Lazy initialization of the global TypeClient instance
+    """
+    global TYPE_CLIENT
+    if TYPE_CLIENT is None:
+        conn = HttpHook.get_connection('search_api_connection')
+        if conn.port is None:
```
mruffalo, line 1063: Is it `conn.conn_type` or `conn.type`? There's a mismatch between the `if`/`else` branches.

To eliminate that possibility, maybe something like:
```python
port_str = '' if conn.port is None else f':{conn.port}'
url = f'{conn.type}://{conn.host}{port_str}'
```

(with either `conn.type` or `conn.conn_type` as appropriate)

mruffalo (changes requested):

src/ingest-pipeline/airflow/dags/utils.py
jswelling: Ooh, good catch, thanks!

PR #346, opened 2021-06-15
Change lightsheet regex in workflow_map to match a broader range of s…

jswelling:
…trings

Until assay type names get fully standardized, we need to recognize metadata assay_type entries like 'Light Sheet' and 'lightsheet' as both meaning the same assay type.  This PR modifies a regex in workflow_map.yml to make that change.


icaoberg (approved):
Looks correct 👍

Issue #323, opened 2021-05-21
Provide a service ingest-api can use to validate an upload


jswelling:
This could be a modified version of the existing call.  It needs to:
* provide a workflow name for the new facility
* the call will include an 'upload' rather than a 'dataset'
* after testing, the run needs to return the results via an API call on entity-api to `https://entity-api/entities/<upload uuid>` with payload
  ```
      {
          "status":"Invalid",
          "validation_message":"This Upload is invalid because duis eu justo lorem. Sed euismod nunc orci. Nunc ornare massa a enim pellentesque."
      }
  ``` 
  Where status can be one of: Invalid, Error or Valid, if Valid status is returned "validataion_message" is optional, required otherwise.

jswelling:
The part of app.py making the incoming call is: https://github.com/hubmapconsortium/ingest-api/blob/9a22f8724af0d4685c1723cb743feabb9619a926/src/app.py#L1377-L1401

khanshawPSC:
Waiting on a PR review

PR #353, opened 2021-06-21
Welling/add validate endpoint

jswelling:
Changes to hubpmap_api plugin endpoint.py and new validate_upload.py DAG to permit ingest-api to trigger execution of ingest-validation-tools and ingest-validation-tests and to return results.  Includes bumping hubmap_api's API version number to 4. 

Addresses https://github.com/hubmapconsortium/ingest-pipeline/issues/323


jswelling:
This still requires:
* demonstration that it can be called
* resolve callback cross-site token issue
* demonstrate correct behavior under Valid, Invalid, Error failure modes.

PR #357, opened 2021-06-24
Welling/trigger downstream processing

jswelling:
Add a dag that can be manually triggered, which leaves the parent dataset unchanged and will start the downstream processing specified in workflow_map.yml regardless of the status of the parent dataset.


khanshawPSC (approved):
No changes necessary

Issue #328, opened 2021-06-03
Register Dataset thumbnail files at data ingest


shirey:
For all datasets if the file, extras/thumbnail.jpg exists register it as a file to be published via the assets service.

yuanzhou:
Notes: 
- For primary dataset, not the secondary
- It happens when the Submit button is hit (the dataset metadata information is already created), not through the ingest UI though. The `extras/thumbnail.jpg ` will be uploaded via Globus with the data files. 
- I'll need to look for `extras/thumbnail.jpg` and register it then commit it. Will decide the workflow.
- Create a new branch off `devel` and send a PR back to `devel` with Joel as the reviewer. 
- Code: https://github.com/hubmapconsortium/ingest-pipeline/blob/devel/src/ingest-pipeline/airflow/dags/utils.py#L934
- Test the work against Airflow on STAGE

yuanzhou:
Updates:
- Use `dataset_lz_path_fun` (https://github.com/hubmapconsortium/ingest-pipeline/blob/devel/src/ingest-pipeline/airflow/dags/utils.py#L875) to get the full path to the dataset
- Check the existence of `extra/thumbnail.jpg` under the dataset  full path
- Add the full thumbnail file path to something like `md['thumbnail_path'] = path` to be sent back to ingest-api
- `Dataset.get_dataset_ingest_update_record()` (https://github.com/hubmapconsortium/ingest-api/blob/790adaa789a6378b4385d79e6e6bd78ba10023f2/src/dataset.py#L194) parses the resulting data from ingest-pipeline and send the updated dataset record to entity-api to store the information in Neo4j
- Add a new property in entity-api schema yaml for `Dataset.thumbnail_image` with trigger methods to get back the tmp id then commit the file to get back file_uuid. The copy of this thumbnail image will be under the `hive/hubmap/hm_uploads/<dataset_uuid>/<file_uuid>`
- The ingest-api will also need to create a symbolic link from this `hive/hubmap/hm_uploads/<dataset_uuid>/<file_uuid>` to the assets `hive/hubmap/assets/<file_uuid>` but without the dataset_uuid directory to avoid conflicts.
- Then this thumbnail file will be accessible via the assets file service. Token is needed if the dataset is consortium, public dataset doesn't require a token.

PR #367, opened 2021-07-02
Register thumbnail image

yuanzhou:
This PR addresses issue #328  with the following workflow:

- ingest-pipeline looks for the `extra/thumbnail.jpg` for a given dataset, and send back the absolute file path if found
- ingest-pipeline makes a PUT call to ingest-api's `/datasets/status` endpoint with the metadata
- ingest-api looks for the `thumbnail_file_abs_path` field and handles the dataset update along with all other updated records via entity-api.

@jswelling I've deployed all the related changes from entity-api/gateway/ingest-api on STAGE and tested the subsequent handling without using Airflow. I'm not sure if the changes I made in this PR would work, so can you please review and if all good, can you merge and deploy to STAGE so we can try it out?

### DIFF
--- a/src/ingest-pipeline/airflow/dags/utils.py
+++ b/src/ingest-pipeline/airflow/dags/utils.py
@@ -936,9 +936,10 @@ def make_send_status_msg_function(
             ###################################################################################
             # Added by Zhou 6/16/2021 for registering thumbnail image
             dataset_dir_abs_path = dataset_lz_path_fun(**kwargs)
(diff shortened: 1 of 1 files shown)


jswelling (approved):
I think this is a good interim solution.  I believe there is a separate discussion being had about what the 'appropriate' directory for the thumbnail images would be, with the understanding that 'extras' is not it, but settling on 'extras' for now.

PR #368, opened 2021-07-13
Fix incorrect thumbnail path

yuanzhou:
Should be `extras/thumbnail.jpg` instead of `extra/thumbnail.jpg`


jswelling (approved):
Yup, I didn't catch that either.

PR #385, opened 2021-08-06
CODEX pipeline: use CPU/GPU sub-workflows

mruffalo:
This splits illumination correction, stitching, and background subtraction into separate Airflow operators, so these steps can be run on general-purpose compute nodes and not GPU nodes. Cytokit is now the only task/operator that specifies a GPU node.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/codex_cytokit.py
+++ b/src/ingest-pipeline/airflow/dags/codex_cytokit.py
@@ -51,8 +51,11 @@ with DAG('codex_cytokit',
          ) as dag:
 
     pipeline_name = 'codex-pipeline'
+    steps_dir = Path(pipeline_name) / 'steps'
     cwl_workflows = get_named_absolute_workflows(
-        cytokit=Path(pipeline_name, 'pipeline.cwl'),
+        illumination_first_stitching=steps_dir / 'illumination_first_stitching.cwl',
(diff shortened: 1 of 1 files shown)


src/ingest-pipeline/airflow/dags/codex_cytokit.py
```diff
@@ -488,17 +604,67 @@ def build_cwltool_cmd_sprm_to_anndata(**kwargs):
...
+            >> t_build_cmd_sprm_to_anndata
+            >> t_pipeline_exec_cwl_sprm_to_anndata
+            >> t_maybe_keep_cwl_sprm_to_anndata
+
+            >> t_move_data
+            >> t_expand_symlinks
+            >> t_send_status
+            >> t_join
+    )
+    t_pipeline_exec_cwl_illumination_first_stitching >> t_delete_internal_pipeline_files
```
mruffalo, line 664: I'm not quite sure about this in particular -- my intention is that files which weren't previously exposed from the pipeline (only passed between steps) will be deleted from disk after the pipeline completes. I'm not sure if this is the right way to trigger that, though.

jswelling: That's interesting- I think that should work.  Let's give it a try.

jswelling (approved):

PR #434, opened 2021-10-19
Welling/survey upload support

jswelling:
Adds support for Uploads to survey.py and the utilities like split_and_create that use it.

### DIFF
--- a/src/ingest-pipeline/misc/tools/survey.py
+++ b/src/ingest-pipeline/misc/tools/survey.py
@@ -224,16 +224,13 @@ class Dataset(Entity):
     @property
     def donor_uuid(self):
         if self._donor_uuid is None:
-            for parent_uuid in self.parent_uuids:
-                parent_entity = self.entity_factory.get(parent_uuid)
-                if isinstance(parent_entity, Donor):
-                    self._donor_uuid = parent_uuid
-                    break
-                else:
-                    parent_donor_uuid = parent_entity.donor_uuid
-                    if parent_donor_uuid is not None:
-                        self._donor_uuid = parent_donor_uuid
-                        break
+            parent_uuid = self.parent_uuids[0]
+            parent_entity = self.entity_factory.get(parent_uuid)
+            if isinstance(parent_entity, Donor):
+                self._donor_uuid = parent_uuid
+            else:
+                parent_donor_uuid = parent_entity.donor_uuid
+                self._donor_uuid = parent_donor_uuid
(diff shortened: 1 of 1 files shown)


src/ingest-pipeline/misc/tools/survey.py
```diff
@@ -198,15 +211,46 @@ def __init__(self, prop_dct, entity_factory):
...
         if isinstance(c_h_g, str):
             c_h_g = (c_h_g.lower() in ['yes', 'true'])
         self.contains_human_genetic_sequences = c_h_g
         self._kid_dct = None
         self._parent_dct = None
-    
+        self._organs = None
+
+    @property
+    def donor_uuid(self):
```
jpuerto-psc, line 225: I don't quite understand the difference in this if/else case (probably b/c I don't know how this data is structured). Is this meant to be able to handle the case where a parent entity is not marked as a Donor, but does have a donor_uuid associated with it?

src/ingest-pipeline/misc/tools/survey.py
```diff
@@ -325,15 +376,164 @@ def all_dataset_uuids(self):
...
+        if self._organ is None:
+            # There should only be one parent, another Sample
+            for parent_uuid in self.parent_uuids:
+                parent_entity = self.entity_factory.get(parent_uuid)
+                assert isinstance(parent_entity, Sample), 'Outermost sample has no organ code?'
+                self._organ = parent_entity.organ
+        return self._organ
+
+    @property
+    def donor_uuid(self):
```
jpuerto-psc, line 508: Won't this always break on the first parent_uuid in the list?

src/ingest-pipeline/misc/tools/survey.py
jswelling: Ah.  That's awkwardly coded, but the point is,
* if the parent is a Donor, it's uuid is the thing we want.
* if not, the parent is a Dataset or a Sample, in which case evaluating parent_entity.donor_uuid (with no leading underscore) will recursively call its donor_uuid property, causing it to check *its* parent.  So it's recursion with a property call each time, until it reaches the Donor and the first clause executes.  But we fill in all the _donor_uuid properties on the way down, so they are available in the cache.

src/ingest-pipeline/misc/tools/survey.py
jswelling: Yeah, I agree :-/ .  I'll restructure that.

PR #452, opened 2021-11-08
Welling/utils api no conn

jswelling:
Purpose:
* Modify utils.pythonop_get_dataset_state() to provide an additional entry in its output dict if the given uuid is that of a Dataset, containing a list of organ codes like "organs":['LI']
* Remove http_conn_id from the parameters of utils.pythonop_get_dataset_state(), because the organ info must be grabbed from a different connection.
* remove the http_conn_id keyword from the various DAGs that call that routine

### DIFF
--- a/src/ingest-pipeline/airflow/dags/utils.py
+++ b/src/ingest-pipeline/airflow/dags/utils.py
@@ -784,7 +784,7 @@ def pythonop_get_dataset_state(**kwargs) -> JSONType:
         'content-type' : 'application/json',
         'X-Hubmap-Application' : 'ingest-pipeline',
         }
-    http_hook = HttpHook(method, http_conn_id='ingest-api-connection')
+    http_hook = HttpHook(method, http_conn_id='ingest_api_connection')
 
     endpoint = f'entities/{uuid}'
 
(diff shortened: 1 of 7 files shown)


src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -845,6 +844,27 @@ def pythonop_get_dataset_state(**kwargs) -> JSONType:
...
+            print('organs_query_rslt:')
+            pprint(organs_query_rslt)
+        except HTTPError as e:
+            print(f'ERROR: {e}')
+            if e.response.status_code == codes.unauthorized:
+                raise RuntimeError('entity database authorization was rejected?')
+            else:
+                print('benign error')
+                return {}
+        rslt['organs'] = [entry['organ'] for entry in organs_query_rslt]
```
jpuerto-psc, line 866: might be best to move this inside of the try: statement, since organs_query_rslt is scoped there.

jswelling: Done, though the statement was unreachable if the try: failed.

jpuerto-psc (approved):

PR #453, opened 2021-11-11
Hartman/azimuth annotate with updates

jswelling:
This PR adds an azimuth-annotate step to the salmon_rnaseq family of DAGs

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -43,3 +43,6 @@
 [submodule "src/ingest-pipeline/airflow/dags/cwl/celldive-pipeline"]
 	path = src/ingest-pipeline/airflow/dags/cwl/celldive-pipeline
 	url = https://github.com//hubmapconsortium/celldive-pipeline
+[submodule "src/ingest-pipeline/airflow/dags/cwl/azimuth-annotate"]
+	path = src/ingest-pipeline/airflow/dags/cwl/azimuth-annotate
+	url = git@github.com:hubmapconsortium/azimuth-annotate.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -45,4 +45,4 @@
 	url = https://github.com//hubmapconsortium/celldive-pipeline
 [submodule "src/ingest-pipeline/airflow/dags/cwl/azimuth-annotate"]
 	path = src/ingest-pipeline/airflow/dags/cwl/azimuth-annotate
-	url = git@github.com:hubmapconsortium/azimuth-annotate.git
+	url = https://github.com/hubmapconsortium/azimuth-annotate
--- a/src/ingest-pipeline/airflow/dags/salmon_rnaseq.py
+++ b/src/ingest-pipeline/airflow/dags/salmon_rnaseq.py
@@ -111,17 +111,19 @@ def generate_salmon_rnaseq_dag(params: SequencingDagParameters) -> DAG:
             print("tmpdir: ", tmpdir)
 
             # get organ type
-            my_callable = lambda **kwargs: uuid
             ds_rslt = pythonop_get_dataset_state(
-                dataset_uuid_callable=my_callable,
-                http_conn_id='ingest_api_connection',
+                dataset_uuid_callable=get_dataset_uuid,
                 **kwargs
             )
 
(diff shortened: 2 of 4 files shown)


jswelling:
Some mods to the azimuth-annotate docker container are still pending, so review and merging are deferred for now.

AustinHartman (approved):
Looks great, thanks for all your work on this!

mruffalo:
I'm a little confused about commits like 9fa9405. According to the GitHub diff view, that submodule update only modifies code -- code which is baked into a Docker image that's executed by `cwltool`. There's no corresponding update to a newer Docker image though; the tags in both CWL files in https://github.com/hubmapconsortium/azimuth-annotate/tree/main/steps reference image version `1.0`.

@AustinHartman is there a new Docker image corresponding to https://github.com/hubmapconsortium/azimuth-annotate/commit/c2c4f4597a8376c8d6bd07c81bd7d5ded37f201b? Since the tag is still `1.0`, someone could have an old image stored locally, update the pipeline Git repository, and erroneously run the old code in their copy of `hubmap/azimuth-annotate:1.0`.

I see a push to the `1.0` tag yesterday, and the first reference to `hubmap/azimuth-annotate:1.0` is from 2021-11-02, so it looks like this tag is continuously updated.

I _strongly_ recommend that the contents of a Docker image tag be immutable once published, with corresponding tags in the `azimuth-annotate` repository referencing those Docker tags in `steps/*.cwl`. The automation script at https://github.com/hubmapconsortium/pipeline-release-mgmt works very well for this, except for assuming that the main branch name is `master`. This can be overridden with `--main-branch=main` but that isn't yet the default and I also haven't yet added a mechanism for this to be configured in a per-repository configuration file or similar.

mruffalo (changes requested):
I'm a little confused about commits like 9fa9405. According to the GitHub diff view, that submodule update only modifies code -- code which is baked into a Docker image that's executed by `cwltool`. There's no corresponding update to a newer Docker image though; the tags in both CWL files in https://github.com/hubmapconsortium/azimuth-annotate/tree/main/steps reference image version `1.0`.

@AustinHartman is there a new Docker image corresponding to https://github.com/hubmapconsortium/azimuth-annotate/commit/c2c4f4597a8376c8d6bd07c81bd7d5ded37f201b? Since the tag is still `1.0`, someone could have an old image stored locally, update the pipeline Git repository, and erroneously run the old code in their copy of `hubmap/azimuth-annotate:1.0`.

I see a push to the `1.0` tag yesterday, and the first reference to `hubmap/azimuth-annotate:1.0` is from 2021-11-02, so it looks like this tag is continuously updated.

I _strongly_ recommend that the contents of a Docker image tag be immutable once published, with corresponding tags in the `azimuth-annotate` repository referencing those Docker tags in `steps/*.cwl`. The automation script at https://github.com/hubmapconsortium/pipeline-release-mgmt works very well for this, except for assuming that the main branch name is `master`. This can be overridden with `--main-branch=main` but that isn't yet the default and I also haven't yet added a mechanism for this to be configured in a per-repository configuration file or similar.

This would only involve adding a `docker_images.txt` file and modifying the CWL steps to reference `latest` images, then a run of `tag_release_pipeline --main-branch=main ...` to publish a synced version. I'm happy to help with that if appropriate.

AustinHartman:
Thanks, @mruffalo - that's a good point about the code being baked into the docker image, so updating the submodule doesn't matter. It is at least probably useful to point to the submodule commit used to build the docker image. For commit c2c4f45 a new docker image was pushed.

I definitely won't continuously update the image once this is merged. I anticipated only needing to push once, and then changes were requested so I had to push updates on a couple of occasions. I probably should have used some sort of beta versioning to avoid overwriting the pre-release image instead. Thanks for pointing me to the automation script - I'll look into using it going forward.

mruffalo:
@AustinHartman Makes sense -- in that case I'd add a `v1.0` tag in the `azimuth-annotate` repository pointing to https://github.com/hubmapconsortium/azimuth-annotate/commit/c2c4f4597a8376c8d6bd07c81bd7d5ded37f201b, so it's clear that this version corresponds to what's merged and run here.

Thanks, and I'm looking forward to seeing the results across all data sets!

mruffalo (approved):

PR #469, opened 2021-11-30
Bump sprm submodule to v1.0.6.3

mruffalo:
Fixes a few issues, including another occurrence of a failure in PCA decomposition. @tCeZ can elaborate if desired.


tCeZ:
Anywhere there is a need to solve for full svd of a full rank matrix - put into try/catch block to further reduce dataset if necessary. Also changed how adj matrix is calculated and shown.

jswelling (approved):

PR #487, opened 2022-01-20
Welling/reorganize upload automation

jswelling:
This PR adds the reorganize_upload DAG, and adds logic to the HuBMAP API plug-in to call that DAG.  split_and_create.py also gets some updates, because functionality from that module must be imported by the new DAG.


jswelling:
The branch diverged from devel; I need to update the branch to avoid a messy merge.

jpuerto-psc (approved):

PR #548, opened 2022-04-06
Welling/unify resource control

jswelling:
This PR adds mechanisms to control resource usage (queue, threads, 'lanes' or simultaneous DAG runs) from a central file, resource_map.yml .  It includes:
* definition and schema for resource_map.yml
* routines in utils to access those values: get_queue_resource(dag_id), etc.
* a derived class of DAG, utils.HMDAG, that automatically applies the queue and lanes resources when a DAG is created
* mods to all existing DAG code to use HMDAG instead of DAG, and thus to use the new resource table.

To understand it, look at the changes in utils.py, and then look at some dag code like salmon_rnaseq.py to understand how it is used.


jswelling:
The merge conflicts aren't 'real', except for reorganize_upload which I will fix at merge time.

src/ingest-pipeline/airflow/dags/resource_map.yml
```diff
@@ -0,0 +1,76 @@
...
+      'queue': 'gpu000_q1'
+      'threads': 6
+    - 'task_re': '.*'
+      'queue': 'general'
+      'threads': 6
+  - 'dag_re': 'ometiff_offset'
+    'lanes': 2
+    'tasks':
+    - 'task_re': '.*_exec_cwl1'
+      'queue': 'gpu000_q1'
```
mruffalo, line 27: As far as I remember, the OME-TIFF offsets calculation doesn't require a GPU node; the CWL workflow is executed without any `--gpus` argument.

mruffalo, line 27: As far as I remember, the OME-TIFF offsets calculation doesn't require a GPU node; the CWL workflow is executed without any `--gpus` argument.

src/ingest-pipeline/airflow/dags/salmon_rnaseq.py
```diff
@@ -30,11 +30,11 @@
...
+    HMDAG,
+    get_queue_resource,
+    get_threads_resource,
 )
 
-# to be used by the CWL worker
-THREADS = 6
-
 
 def generate_salmon_rnaseq_dag(params: SequencingDagParameters) -> DAG:
```
mruffalo, line 39: I'm not sure how to add a comment to the `salmon-rnaseq` submodule update, so I'm adding it here -- is that spurious? The GitHub diff view doesn't show any relevant changes.

src/ingest-pipeline/airflow/dags/resource_map.yml
```diff
@@ -0,0 +1,76 @@
...
+      'threads': 6
+    - 'task_re': '.*'
+      'queue': 'general'
+      'threads': 6
+  - 'dag_re': 'sc_atac_seq_.*'
+    'lanes': 8
+    'tasks':
+    - 'task_re': '.*'
+      'queue': 'general'
+      'threads': 6
```
mruffalo, line 70: Could probably increase this quite a bit, depending on resource availability. This would significantly decrease the runtime for scATAC-seq runs.

mruffalo (commented):
Looks good, except for the submodule updates that look slightly out of place.

src/ingest-pipeline/airflow/dags/salmon_rnaseq.py
jswelling: I apparently updated salmon-rnaseq to match the current version on devel to run a test.  That was different from the version in effect when the branch was created, so it showed up as a change when I created the PR and I failed to notice.  I've cleaned it up; devel already has the change in place.

src/ingest-pipeline/airflow/dags/resource_map.yml
jswelling: fixed.

PR #549, opened 2022-04-07
Always use permanent tok when an airflow workflow is triggered. WIP.

jswelling:
This PR substitutes the 'permanent token' for the token of the UI user when initiating Airflow runs via the Airflow plug-in endpoint.


shirey (approved):
Looks good to me.

PR #587, opened 2022-05-17
Welling/add bdbag dag

jswelling:
This adds the equivalent of py-hubmapbags/hubmapbags/magic do_it() as a DAG to run under Airflow.


icaoberg (approved):
Looks good but if I update the package, how can you update this code as well?

PR #597, opened 2022-06-02
Welling/scan bail on failed validate

jswelling:
This PR modifies the Airflow DAG used to ingest primary datasets, such that a cleaner exit occurs if the dataset fails validation.  The previous state of the DAG caused some extraneous messages to be included in the diagnostics uploaded as a result of the failure.  This change prevents those extraneous messages and thus prevents their inclusion in the diagnostic presented to the data provider when a dataset is found to be invalid.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
+++ b/src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
@@ -184,7 +184,8 @@ with HMDAG('scan_and_begin_processing',
     t_send_status = PythonOperator(
         task_id='send_status_msg',
         python_callable=wrapped_send_status_msg,
-        provide_context=True
+        provide_context=True,
+        trigger_rule='all_done'
     )
 
     t_create_tmpdir = BashOperator(
@@ -192,13 +193,11 @@ with HMDAG('scan_and_begin_processing',
         bash_command='mkdir {{tmp_dir_path(run_id)}}'
         )
 
-
     t_cleanup_tmpdir = BashOperator(
         task_id='cleanup_temp_dir',
         bash_command='echo rm -r {{tmp_dir_path(run_id)}}',
         trigger_rule='all_success'
         )
-
 
     def flex_maybe_spawn(**kwargs):
         """
@@ -210,8 +209,10 @@ with HMDAG('scan_and_begin_processing',
         ctx = kwargs['dag_run'].conf
         pprint(ctx)
         run_validation_retcode = int(kwargs['ti'].xcom_pull(task_ids="run_validation"))
-        md_extract_retcode = int(kwargs['ti'].xcom_pull(task_ids="run_md_extract"))
-        md_consistency_retcode = int(kwargs['ti'].xcom_pull(task_ids="md_consistency_tests"))
+        md_extract_retcode = kwargs['ti'].xcom_pull(task_ids="run_md_extract")
+        md_extract_retcode = int(md_extract_retcode or '0')
+        md_consistency_retcode = kwargs['ti'].xcom_pull(task_ids="md_consistency_tests")
+        md_consistency_retcode = int(md_consistency_retcode or '0')
         if (run_validation_retcode == 0
             and md_extract_retcode == 0
             and md_consistency_retcode == 0):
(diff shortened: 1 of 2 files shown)


src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
```diff
@@ -199,8 +209,10 @@ def flex_maybe_spawn(**kwargs):
         ctx = kwargs['dag_run'].conf
         pprint(ctx)
         run_validation_retcode = int(kwargs['ti'].xcom_pull(task_ids="run_validation"))
-        md_extract_retcode = int(kwargs['ti'].xcom_pull(task_ids="run_md_extract"))
-        md_consistency_retcode = int(kwargs['ti'].xcom_pull(task_ids="md_consistency_tests"))
+        md_extract_retcode = kwargs['ti'].xcom_pull(task_ids="run_md_extract")
```
Daghis, line 212: I'm curious why you broke this into two statements instead of adding the `or '0'` to lines 202-203 within the `int()` call.

src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
```diff
@@ -233,6 +245,7 @@ def flex_maybe_spawn(**kwargs):
         )
 
     (dag >> t_create_tmpdir >> t_run_validation >>
+     t_maybe_continue >>
```
Daghis, line 248: Educational question:

```dag >> t_create_tmpdir >> t_run_validation >> t_maybe_continue >> t_run_md_extract >> ...```

Technically-speaking only, is this specific connection here between `t_maybe_continue` and `t_run_md_extract` required given that `t_maybe_continue` has `'next_op': 'run_md_extract'` in its definition?

Style and clarity are obviously important, but I just wanted to understand the net effect of one versus the other.

File: src/ingest-pipeline/airflow/dags/utils.py [hunk 3]
```diff
@@ -1111,9 +1111,10 @@ def make_send_status_msg_function(
     """
     def send_status_msg(**kwargs) -> bool:
         retcodes = [
-            int(kwargs['ti'].xcom_pull(task_ids=op))
+            kwargs['ti'].xcom_pull(task_ids=op)
             for op in retcode_ops
         ]
+        retcodes = [int(rc or '0') for rc in retcodes]
```
Daghis, line 1117: Repeating an earlier question, why not have `or '0'` on line 1114 inside the `int()`?

src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
Daghis, line 212: I'm curious why you broke this into two statements instead of adding the `or '0'` to lines 202-203 within the `int()` call.

Daghis (commented):
Seems straightforward enough in terms of this logic. What I can't review is the behavior of inserting `run_validation` into the dag (which is what I think is happening by adding `t_maybe_continue`).

Daghis (approved):

src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
jswelling: One possible code path was causing the xcom_pull to return None, which is an invalid input to int().  Splitting into two lines to guarantee that the input to int() would be a valid string '0' if the return value was None seemed like a fair compromise between being concise and clear.

src/ingest-pipeline/airflow/dags/scan_and_begin_processing.py
jswelling: Yes, it turns out to explicitly needed.  I guess that's a design decision on the part of the Airflow authors.

src/ingest-pipeline/airflow/dags/utils.py
jswelling: That would work- the 'or' clause in the list completion just seemed awkward to me.  Your approach is equivalent, and may be clearer.

Daghis: Thank you for your answers!

PR #600, opened 2022-06-10
Add fastq-utils to requirements.txt

jswelling:
This adds fastq-utils to requirements.txt via github.  Since this is now a requirement of ingest-validation-tests, one of our submodules, it must be a requirement here as well.


Daghis:
Thanks!

PR #620, opened 2022-06-29
Freeze airflow at 1.10.15 in requirements.txt

jswelling:
This PR freezes Airflow at 1.10.15 in requirements.txt .  This has been the deployed version for a while, but leaving it unspecified was leading to undesirable version interpolation by pip when the environment was built.


sunset666 (approved):
Having exact versions on the requirements makes it easy to deploy and prevents errors.

PR #621, opened 2022-06-30
sunset666/issue_619

sunset666:


src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -0,0 +1,88 @@
...
+    parser.add_argument("uuid_txt",
+                        help="input .txt file containing uuids or .csv or .tsv file with uuid column")
+    parser.add_argument("--out", help="name of the output .tsv file", required=True)
+    args = parser.parse_args()
+    auth_tok = input('auth_tok: ')
+    entity_factory = EntityFactory(auth_tok)
+
+    uuid_l = []
+    if args.uuid_txt.endswith((".csv", ".tsv")):
+        in_df = pd.read_csv(args.uuid_txt, engine="python", sep=None, encoding='utf-8-sig')
```
Daghis, line 24: What is `in_df`? I get the "in", but I don't know what "df" is.

src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -0,0 +1,88 @@
...
+        if 'uuid' in in_df.columns:
+            uuid_key = 'uuid'
+        elif 'e.uuid' in in_df.columns:
+            uuid_key = 'e.uuid'
+        else:
+            raise RuntimeError(f'Cannot find uuid column in {args.uuid_txt}')
+        for elt in in_df[uuid_key]:
```
Daghis, line 31: What's `elt`?

Daghis (changes requested):
I'm sending this back to you about halfway through. There are a whole bunch of short variable names that I don't know what they're supposed to be. I'm extremely concerned about the amount of technical debt that this project has already and I'd really much rather avoid adding any more at this point. (We need to dig out of this hole.)

src/ingest-pipeline/misc/tools/extract_for_review.py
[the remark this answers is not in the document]
Daghis: It's minor, but I have to admit I like the way it looks now cleaned up. Thanks!

src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -0,0 +1,73 @@
...
+    out_results = []
+    for uuid in uuid_list:
+        result_list = []
+        try:
+            dataset = entity_factory.get(uuid)
+            if dataset.parents is not None:
+                for parent_uuid, _dataset in dataset.parents.items():
+                    data = {'derived_dataset_uuid': dataset.uuid,
+                            'derived_hubmap_id': dataset.hubmap_id,
+                            'derived_portal_url': 'https://portal.hubmapconsortium.org/browse/dataset/%s' % dataset.uuid,
```
Daghis, line 45: Recommend using an f-string here.

src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -0,0 +1,73 @@
...
+                            'primary_dataset_uuid': parent_uuid,
+                            'primary_hubmap_id': _dataset.hubmap_id,
+                            'primary_dataset_uscs/name': _dataset.prop_dct.get('lab_dataset_id', None),
+                            'primary_portal_url': 'https://portal.hubmapconsortium.org/browse/dataset/%s' % parent_uuid}
```
Daghis, line 49: Recommend using an f-string here.

src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -0,0 +1,73 @@
...
+                    result_list.append(data)
+        except SurveyException as e:
+            print(f"dropping {uuid} because {e}")
+            result_list.append({'derived_dataset_uuid': uuid,
+                                'derived_hubmap_id': f'not in survey because {e}',
```
Daghis, line 54: Probably should be double-quotes for this one.

Daghis (approved):

PR #629, opened 2022-07-13
issue_602

sunset666:
This PR removes a sudo command performed by move task on airflow DAGs, this isn't needed and can mislead when debugging.


Daghis (approved):

jswelling (changes requested):
If this is the only reference to OUTPUT_GROUP_NAME in the config information, that parameter should probably be removed from the example config as well.

jswelling (approved):

jpuerto-psc (approved):

PR #640, opened 2022-08-01
Update utility to extract tabular data to aid TMCs reviewing datasets for publication

sunset666:
Updated script to show the information from the parent dataset on derived when switched from lab_dataset_id to provider_info at ingest time

### DIFF
--- a/src/ingest-pipeline/misc/tools/extract_for_review.py
+++ b/src/ingest-pipeline/misc/tools/extract_for_review.py
@@ -43,7 +43,17 @@ def main():
                     dataset_name = ''
                     if _dataset.prop_dct.get('lab_dataset_id'):
                         dataset_name = _dataset.prop_dct.get('lab_dataset_id')
-                    if dataset.prop_dct.get('provider_info'):
+                    elif _dataset.prop_dct.get('provider_info'):
+                        if dataset_name == '':
+                            dataset_name = _dataset.prop_dct.get('provider_info')
+                        else:
+                            dataset_name += '/' + _dataset.prop_dct.get('provider_info')
+                    if dataset.prop_dct.get('lab_dataset_id'):
+                        if dataset_name == '':
+                            dataset_name = dataset.prop_dct.get('lab_dataset_id')
+                        else:
+                            dataset_name += '/' + dataset.prop_dct.get('lab_dataset_id')
+                    elif dataset.prop_dct.get('provider_info'):
                         if dataset_name == '':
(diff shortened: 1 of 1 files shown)


src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -43,7 +43,17 @@ def main():
                     dataset_name = ''
                     if _dataset.prop_dct.get('lab_dataset_id'):
                         dataset_name = _dataset.prop_dct.get('lab_dataset_id')
-                    if dataset.prop_dct.get('provider_info'):
+                    elif _dataset.prop_dct.get('provider_info'):
+                        if dataset_name == '':
```
Daghis, line 47: I'm pretty sure this condition is always `True`.

src/ingest-pipeline/misc/tools/extract_for_review.py
```diff
@@ -43,7 +43,17 @@ def main():
...
+                            dataset_name = _dataset.prop_dct.get('provider_info')
+                        else:
+                            dataset_name += '/' + _dataset.prop_dct.get('provider_info')
+                    if dataset.prop_dct.get('lab_dataset_id'):
+                        if dataset_name == '':
+                            dataset_name = dataset.prop_dct.get('lab_dataset_id')
+                        else:
+                            dataset_name += '/' + dataset.prop_dct.get('lab_dataset_id')
+                    elif dataset.prop_dct.get('provider_info'):
```
Daghis, line 56: This whole block of code is very confusion. I had to graph out all of the conditionals, plus using `dataset_name` as a conditional as well as value storage made it worse, I'm afraid.

So, after graphing it out, I came up with this skeleton. I believe this code should achieve the same result, but may be more readable. What do you think?

```
if dataset.parents is not None:
    for parent_uuid, parent_dataset in dataset.parents.items():
        parent_component = (
                parent_dataset.prop_dct.get('lab_dataset_id')
                or parent_dataset.prop_dct.get('provider_info')
        )
        child_component = (
                dataset.prop_dct.get('lab_dataset_id')
                or dataset.prop_dct.get('provider_info')
        )
        dataset_name = ' / '.join(c for c in [parent_component,
                                              child_component] if c)
```

Daghis (commented):
May I suggest that the title of the PR should be more descriptive of the intended change. That can make browsing the commit history much easier to find what may have impacted a particular area of code.

Daghis (approved):
Still would request a more descriptive title for the PR.

PR #646, opened 2022-09-08
Integrate PAS FTU segmentation

mruffalo:


sunset666 (changes requested):
Since we launched Airflow 2.x and this was created before that, the code is a bit outdated. Please check the new structure from current version to update it accordingly.

jswelling (approved):

PR #660, opened 2022-10-17
Parallelization on ingest validation test control

sunset666:
New resource_map entry to control how many threads can the new parallel processing on gz validator plugin can make use of.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/utils.py
+++ b/src/ingest-pipeline/airflow/dags/utils.py
@@ -1396,6 +1396,15 @@ def get_threads_resource(dag_id: str, task_id: Optional[str] = None) -> int:
     return int(rec['threads'])
 
 
+def get_core_use_resource(dag_id: str, task_id: Optional[str] = None) -> int:
+    """
+    Look up the number of cores percentage to use as thread for parallel processing
+    """
+    rec = _lookup_resource_record(dag_id, task_id)
+    assert 'coreuse' in rec, 'schema should guarantee "coreuse" is present?'
+    return int(rec['coreuse'])
+
+
 def get_type_client() -> TypeClient:
     """
     Expose the type client instance publicly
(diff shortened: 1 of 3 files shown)


jswelling (changes requested):
I have to disagree with the design of this one, in that the 'theads' entry and the 'coreuse' entry are serving the same purpose.  They shouldn't both be allowed, and they shouldn't both have separate get_*_resource() calls.  Couldn't you do something like 'threads' and 'coreusepercent', but:
* modify the schema so only one of them can occur per entry- this is straightforward, I think
* have utils.py do the logic to count the cores and calculate threads, returning it via get_threads_resource()?  
I like the fact that you found a way to pass the thread count in via extra_parameters rather than an environment variable, but I think the extra_paramter should be just {'threads':...}

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -1392,8 +1392,8 @@ def get_threads_resource(dag_id: str, task_id: Optional[str] = None) -> int:
     if task_id is None:
         task_id = '__default__'
     rec = _lookup_resource_record(dag_id, task_id)
-    assert 'threads' in rec, 'schema should guarantee "threads" is present?'
-    return int(rec['threads'])
+    assert any(['threads' in rec, 'coreuse' in rec]), 'schema should guarantee "threads" or "coreuse" is present?'
+    return int(rec.get('coreuse') if rec.get('coreuse') else rec.get('threads'))
```
jswelling, line 1396: This can't be right, since 'coreuse' might be 60 and we want to return 60% of the cores on the node.  It needs to get converted to a core count.  Also, it has to never return zero, because I think that would break things- I think it should be ceil(float(rec.get('coreuse'))*0.01*number_of_cores)

jswelling (changes requested):

src/ingest-pipeline/airflow/dags/utils.py
sunset666: It is a different approach, currently the calculation is performed on the plugin. But this makes me realize that it will need more coding to make use of this new parameter. I'll refactor it so it is centralized.

jswelling (approved):

PR #666, opened 2022-10-27
Refactoring of new_dataset_survey to connect directly to google docs

sunset666:
Refactoring of new_dataset_survey to connect directly to the google docs spreadsheet, and a workaround for bogus datasets with no parent.

### DIFF
--- a/src/ingest-pipeline/misc/tools/new_dataset_survey.py
+++ b/src/ingest-pipeline/misc/tools/new_dataset_survey.py
@@ -1,15 +1,18 @@
 #! /usr/bin/env python
+import os
 
-import sys
+from googleapiclient.discovery import build
+from google.auth.transport.requests import Request
+from google_auth_oauthlib.flow import InstalledAppFlow
+from google.oauth2.credentials import Credentials
+from googleapiclient.errors import HttpError
+
 import argparse
-from pprint import pprint
 from datetime import date
 import pandas as pd
 import numpy as np
 
-from survey import (Entity, Dataset, Sample, EntityFactory,
-                    ROW_SORT_KEYS, column_sorter, is_uuid,
-                    parse_text_list, SurveyException)
+from survey import (EntityFactory, ROW_SORT_KEYS, column_sorter, is_uuid, parse_text_list, SurveyException)
 
 
 """
@@ -21,6 +24,15 @@ VOLATILE_NOTES = set(['BAD TYPE NAME',
                       'Multiple QA derived datasets',
                       'BAD UUID: No parents?',
                   ])
+
+# If modifying these scopes, delete the file token.json.
+SCOPES = ['https://www.googleapis.com/auth/spreadsheets']
+API_SERVICE_NAME = 'sheets'
+API_VERSION = 'v4'
+
+# The ID and range of a sample spreadsheet.
+# SPREADSHEET_ID = '1-CF0R-rgfmeHMUjLkii7Qv6QbPqMscShJJtLIXB4-74'
+SPREADSHEET_ID = '1xTzQRtG841d1ulSM9k-7m3ss1ms4uXPjCtZnEP2QMKA'
 
 
 def detect_otherdata(ds):
@@ -42,8 +54,8 @@ def detect_metadatatsv(ds):
     for path in ds.full_path.glob('*metadata.tsv'):
         md_df = pd.read_csv(path, sep='\t')
         if 'assay_type' in md_df.columns:
-            return (True, len(md_df))
-    return (False, 0)
+            return True, len(md_df)
+    return False, 0
 
 
 def detect_clean_validation_report(ds):
@@ -75,10 +87,8 @@ def get_most_recent_touch(ds):
 def data_type_resolver(row):
     dt_x = parse_text_list(row["data_types_x"])
     dt_y = parse_text_list(row["data_types_y"])
-    if ((isinstance(dt_x, str) and dt_x.lower() == 'nan')
-        or (isinstance(dt_x, float) and dt_x == np.nan)):
-        if ((isinstance(dt_y, str) and dt_y.lower() == 'nan')
-            or (isinstance(dt_y, float) and dt_y == np.nan)):
+    if (isinstance(dt_x, str) and dt_x.lower() == 'nan') or (isinstance(dt_x, float) and dt_x == np.nan):
+        if (isinstance(dt_y, str) and dt_y.lower() == 'nan') or (isinstance(dt_y, float) and dt_y == np.nan):
             return '????'
         else:
             return f'{dt_y}'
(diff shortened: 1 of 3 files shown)


src/ingest-pipeline/requirements.txt
```diff
@@ -32,4 +32,6 @@ xmlschema>=1.6
 WTForms<=2.3.3
 hubmap_sdk
 pandas
-
+google-api-python-client==2.52.0
```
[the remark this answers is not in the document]
sunset666: we eventually need to fix all those requirements and keep upgrading them accordingly. Leaving them open for PIP to figure them out will cause bigger issues.

src/ingest-pipeline/misc/tools/new_dataset_survey.py
```diff
@@ -249,11 +365,8 @@ def main():
     out_df = out_df.sort_values([key for key in ROW_SORT_KEYS if key in out_df.columns],
                                 axis=0)
 
-    out_df.to_csv(args.out, sep='\t', index=False,
```
jswelling, line 252: I propose that there be a command line option to send directly to Google Sheets, so the old functionality is available if needed.

src/ingest-pipeline/misc/tools/new_dataset_survey.py
```diff
@@ -115,6 +126,114 @@ def join_notes(df, notes_df):
...
+    if os.path.exists('.token.json'):
+        creds = Credentials.from_authorized_user_file('.token.json', SCOPES)
+    # If there are no (valid) credentials available, let the user log in.
+    if not creds or not creds.valid:
+        if creds and creds.expired and creds.refresh_token:
+            creds.refresh(Request())
+        else:
+            flow = InstalledAppFlow.from_client_secrets_file(
+                '.sheetreader_key.json', SCOPES)
+            creds = flow.run_local_server(port=0)
```
jswelling, line 143: What actually happens when this path executes?  How does the interaction with the user occur, and how can we get Airflow to deal with it?

src/ingest-pipeline/misc/tools/survey.py
```diff
@@ -220,7 +221,7 @@ def __init__(self, prop_dct, entity_factory):
 
     @property
     def donor_uuid(self):
-        if self._donor_uuid is None:
+        if self._donor_uuid is None and self.parent_uuids:
```
jswelling, line 224: This change results in donor_uuid being None if self.parent_uuids is false, and it really surprises me that that doesn't break anything.  What comes out on the donor_uuid line in the final spreadsheet when this happens?

src/ingest-pipeline/misc/tools/new_dataset_survey.py
sunset666: Goes to the authorization part of Google Apps, it opens a web page when you click authorize (if you are actually allowed to see the doc) and generates de authentication token based on the sheetreaders_key.json created for this particular project.

jswelling (commented):
Three things are happening with this PR:
* A work-around for a current upstream bug involving the parent uuids of a dataset
* Addition of the ability to directly deliver the resulting spreadsheet to Google Sheets
* Some PEP8 formatting changes 

All are reasonable, and I don't object to all happening in the same PR, but it is a little confusing.  

Is .token.json a high-security thing?  Should we include .token.json and .sheetreader_key.json in the current discussion of puppetization of config files?

src/ingest-pipeline/misc/tools/survey.py
sunset666: For all the data we already have, there is always a parent_uuid which makes the donor_uuid not None. Only the bogus ones are failing into this category where _donor_uuid is none.

sunset666:
.token.json is dynamically created the first time this is executed and the .sheetreader_key.json does have the credentials that allows the use of the Google API for the PSC drive, so I believe this one **is** a high-security thing and they might be included in the Puppet Server as config file.

src/ingest-pipeline/requirements.txt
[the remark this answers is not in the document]
jswelling: Concur.

src/ingest-pipeline/misc/tools/new_dataset_survey.py
jswelling: Different parameters to main?  I'm not sure I follow what you're proposing.  I thought to just add a command line flag like --to_csv to cause the old behavior to occur.  (I presume one can still add additional --notes from the command line)

src/ingest-pipeline/misc/tools/survey.py
jswelling: Yes, and I agree that the problem will only occur until Bill's team fixes the bug in their endpoint, but what does the failure mode look like?  I thought it was going to be necessary to create a special entity that would play a role like <NO-SUCH-ENTITY> in order to keep things running smoothly.

src/ingest-pipeline/misc/tools/new_dataset_survey.py
sunset666: Well, the behavior and parameter did change, now you don't use paths for it (--out and --notes). --out now uses the name of the new sheet you want and --notes was updated to --sheet_name (which in retrospective isn't a good name for it). So, my counter-proposal is to rename parameters so they are more adequate to the function you require like this. --to_csv --out_file --notes_file and --out_sheet --notes_sheet (for the new functionality)

src/ingest-pipeline/misc/tools/survey.py
sunset666: It would just skip most of the information for the bogus entities and keep on processing the good ones (but saves enough to be showed in the sheet if you want to identify was is going on with them), you can't actually create a 'special' entity, because that would need to exist in the DB so the entity-search can retrieve something for it. Not sure if it is worth the extra effort. Did I missed something there?

src/ingest-pipeline/misc/tools/new_dataset_survey.py
jswelling: That makes perfect sense- I like those names.

src/ingest-pipeline/misc/tools/new_dataset_survey.py
jswelling: Understood, though I don't see how it's going to open a web page if it's running on hivevm191.

src/ingest-pipeline/misc/tools/survey.py
jswelling: My thought was to insert the special entity in the cache of entity-search results so it wouldn't have to go all the way to entity-api, but I agree that it's not necessary.

src/ingest-pipeline/misc/tools/new_dataset_survey.py
sunset666: You are right, it will give you a link for that but the callback won't work. It will need to be created locally and then moved over the server (the file will be .token.json).

jswelling (approved):

PR #682, opened 2022-11-19
Script to regenerate conda environment and associated issues

jswelling:
This PR addresses the following:
* Add a conda equivalent of the script regenerate_venv.sh
* Put the conda environment in /opt/environments
* Some common code between regenerate_conda_env and airflow_wrapper is moved to a separate script
Testing was performed under Airflow2 on DEV under both python3.6 and python3.9

### DIFF
--- a/src/ingest-pipeline/misc/tools/airflow_wrapper.sh
+++ b/src/ingest-pipeline/misc/tools/airflow_wrapper.sh
@@ -84,13 +84,15 @@ for varname in "${envvars[@]}" ; do
 done
 
 if [ "${HM_AF_METHOD}" == 'conda' ] ; then
-    export PATH=/hive/users/hive/anaconda3/bin:%PATH
-    export LANG=C.UTF-8
-    export LC_ALL=C.UTF-8
+    which conda || export PATH=/hive/users/hive/anaconda3/bin:$PATH
+    #export LC_ALL=C.UTF-8
+    #export LANG=C.UTF-8
     eval "$(conda shell.bash hook)"
     conda activate "${HM_AF_ENV_NAME}"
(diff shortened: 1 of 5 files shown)


src/ingest-pipeline/misc/tools/airflow_wrapper.sh
```diff
@@ -84,13 +70,15 @@ for varname in "${envvars[@]}" ; do
 done
 
 if [ "${HM_AF_METHOD}" == 'conda' ] ; then
-    export PATH=/hive/users/hive/anaconda3/bin:%PATH
-    export LANG=C.UTF-8
-    export LC_ALL=C.UTF-8
+    which conda || export PATH=/hive/users/hive/anaconda3/bin:$PATH
+    #export LC_ALL=C.UTF-8
```
jswelling, line 74: @sunset666 I notice you added the LC_ALL and LANG exports as a bugfix, but my experiments on DEV showed they actually caused errors during start-up.  I'm concerned that the need for these lines may be host-dependent.

jswelling:
Closes https://github.com/hubmapconsortium/ingest-pipeline/issues/671

src/ingest-pipeline/misc/tools/airflow_wrapper.sh
sunset666, line 74: can be removed from the script

src/ingest-pipeline/misc/tools/airflow_wrapper.sh
```diff
@@ -84,13 +70,15 @@ for varname in "${envvars[@]}" ; do
...
+    #export LANG=C.UTF-8
```
sunset666, line 75: can be removed from the script

sunset666 (approved):

PR #685, opened 2022-11-29
Plugin to find error strings in session.log

gesinaphillips:
This PR contains a plugin for the diagnose_failure DAG. This plugin uses regex to search for error strings and logs the lines with errors from session.log.

Note 1: The find_errors_plugin currently contains an override of FindErrorsDiagnosticResult.problem_found that may be unnecessary pending potential changes to parent class.

Note 2: I will name my branch better next time!

### DIFF
--- a/src/ingest-pipeline/airflow/dags/diagnostics/plugins/find_errors_plugin.py
+++ b/src/ingest-pipeline/airflow/dags/diagnostics/plugins/find_errors_plugin.py
@@ -1,0 +1,57 @@
+"""
+This plugin uses regex to search for errors in session.log
+"""
+
+import re
+from pathlib import Path
+
+from diagnostics.diagnostic_plugin import (
+    DiagnosticPlugin,
+    DiagnosticResult,
+    DiagnosticError,
+)
+
+
+# TODO: maybe there's a way to encode case insensitivity directly into the regex string
+# rather than doing it in the combine method?
+
+# Finds all instances of the word "error"
+FIND_ERROR_REGEX = r"(\S+)?error(\S+)?"
+# Finds OSError and BrokenPipeError (test--TODO: needs better modularizing)
+FIND_IO_ERROR_REGEX = r"(oserror)|(brokenpipeerror)"
+
+
+class FindErrorsDiagnosticResult(DiagnosticResult):
+    def problem_found(self):
+        if len(self.string_list) > 0:
+            return True
+        return False
+
+    def to_strings(self):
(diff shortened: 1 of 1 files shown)


gesinaphillips:
Oops, I requested review which changed this from a draft PR to a real PR, apologies!

src/ingest-pipeline/airflow/dags/diagnostics/plugins/find_errors_plugin.py
```diff
@@ -0,0 +1,41 @@
...
+    DiagnosticPlugin,
+    DiagnosticResult,
+)
+
+
+# TODO: maybe there's a way to encode case insensitivity directly into the regex string
+# rather than doing it in the combine method? That is, if we want case insensitivity
+
+# Finds all instances of the word "error"
+FIND_ERROR_REGEX = r"(\S+)?error(\S+)?"
```
jswelling, line 18: I'd probably drop the FIND_ from the beginnings of these constant names, since shorter is better if it doesn't lead to ambiguity.  
It looks like you *can* specify case insensitivity in the regex itself- I hadn't known this but the following should work:
r"(\S+)?(?i)?error(?-i)(\S+)?"
See the description of (?aiLmsux) in the python regex documentation.  Putting the case insensitivity in the regex would allow you to put the proper cases in the OS_ERROR_REGEX which would probably be good.
I'm not completely sure about the (\S+) parts, which match one or more non-whitespace characters.  Given them, I think I was wrong about the string matching "no errors found", because the leading (\S+) would reject the blank between "no" and "errors".  In fact, I don't think your ERROR_REGEX will actually match "look at this error here" - try it!

jswelling (changes requested):

src/ingest-pipeline/airflow/dags/diagnostics/plugins/find_errors_plugin.py
gesinaphillips: I'll try it! My experimentation had it capturing just "error" but I will make sure.

With some reflection, I think I'm trying to do two things at once: 1) specifically capture the full word (e.g. OSError) and 2) capture context. In Case 1 I agree that \S+ is probably too greedy because it will capture punctuation. In Case 2, I am outputting the entire line anyway so the context is captured that way. I will revise the regex accordingly, taking the case in/sensitivity into account as well. Thanks!

jswelling (approved):

PR #690, opened 2022-12-13
export_and_backup DAG

gesinaphillips:
This is the framework for a DAG that runs a set of plugins for a given UUID based on the current status of the dataset (error, published, etc.). It is based largely on the recent diagnose_failure DAG, although the functionality has been simplified for now. There are probably ways I have simplified it that I will need to walk back later, and things I have borrowed that are irrelevant.

Statuses are mapped to appropriate plugins in export_and_backup_map.yml. 

When this DAG runs in its current state with the config {"uuid":"a6ead319af30ab63bee443396d1a031f"} -- a dataset with the status "error" -- it should run the error_backup and error_export plugins, but not the published_backup plugin. This has been tested on DEV.

A number of things are incomplete:
- The set of possible dataset statuses is currently mocked.
- The set of dataset statuses that will be relevant to this process is uncertain (e.g. do we want to backup data that is in an invalid state or that has been abandoned? if so, where/how?).
- This is handwaving uploads vs. datasets currently.


src/ingest-pipeline/airflow/dags/export_and_backup.py
```diff
@@ -0,0 +1,140 @@
...
+    "email": ["joel.welling@gmail.com"],
+    "email_on_failure": False,
+    "email_on_retry": False,
+    "retries": 1,
+    "retry_delay": timedelta(minutes=1),
+    "xcom_push": True,
+    "queue": get_queue_resource("validation_test"),
+}
+
+with HMDAG(
```
sunset666, line 37: Just to conserve the same code structure, please modify the spacing to be the same as all other DAGs (at least previous ones, I haven't looked the new plugin framework)

sunset666 (changes requested):

src/ingest-pipeline/airflow/dags/export_and_backup.py
gesinaphillips: Thanks Sunset! I see that my indentation for lines 36-40 doesn't match other DAGs. I think I also have some inappropriate double-line breaks. I have made these updates, in addition to switching to single-quotes per the other DAGs. Please let me know if you noticed anything else or if there are other changes you would like to request.

I see that we have some rules for linting using black in the repo's [pyproject.toml](https://github.com/hubmapconsortium/ingest-pipeline/blob/devel/pyproject.toml) file. If we want to enforce this style then I think we should update that file (and maybe even add some pre-commit hooks?). I'd be happy to work on figuring out some linting rules. What do you think?

sunset666: I honestly haven't checked that, but any improvement is welcome. If you feel comfortable enough working on it and see a good outcome of it, go for it and teach us.

gesinaphillips: Sunset, @jpuerto-psc, @jswelling, are you all using isort & black? Trying to figure out if my configuration is busted, if we're not all using the same tools, or something else!

jswelling (changes requested):
I think you're going to need to distinguish between the status strings which are legit from Uploads and those which are legit for Datasets.  For example, 'submitted' and 'validated' are legit status strings for Uploads but not for Datasets.
I also think you should check the map against the schema you've provided for it when you read it in.  You can steal boilerplate for that from existing code.  For example see localized_assert_json_matches_schema() in utils.py .
I also think the plugins will want info to help locate the dataset, and that you could get that from the value of ds_rslt you are getting with your xcom_pull requests.

gesinaphillips:
@jswelling I think I made requested changes--please see the updated schema as well as how export_and_backup.py and export_and_backup_map.yml handle uploads vs. datasets. This has been tested on DEV with the changes and successfully reached plugins for datasets with statuses error and published and for upload with status reorganized. I believe the info dict is being passed along appropriately to be reached by plugins but this will require writing and testing plugins (I have some in the pipeline but I will want to make sure they branch from this code when it is merged).

src/ingest-pipeline/schemata/export_and_backup_map_schema.yml
```diff
@@ -0,0 +1,25 @@
...
+
+  'export_and_backup_record':
+     'type': 'object'
+     'properties':
+        'status':
+           'type': 'string'
+           'description': 'the status of the dataset or upload'
+        'plugins':
+            'type': 'array'
+            'description': 'array of appropriate plugins for status'
```
jswelling, line 18: I think add the following to specify the type of items in the plugins array:
'items': {'type': 'string'}

src/ingest-pipeline/schemata/export_and_backup_map_schema.yml
```diff
@@ -0,0 +1,25 @@
...
+  'export_and_backup_entity_type':
+      'type': 'string'
+      'description': 'contains associated export_and_backup_records'
+  'export_and_backup_map':
+    'type': 'object'
+    'properties':
+      'export_and_backup_entity_type': {'$ref': '#/definitions/export_and_backup_entity_type'}
```
jswelling, line 25: This bit doesn't seem right; it's too self-referential.  I think you need something like:
```
'type': 'object'
'properties':
  'dataset':
    'type': 'array'
    'items': {'$ref': '#/definitions/export_and_backup_record'}
    'description': 'plugins applicable to datasets'
  'upload':
    'type': 'array'
    'items': {'$ref': '#/definitions/export_and_backup_record'}
    'description': 'plugins applicable to uploads'

```

jswelling: Schemata are really tricky.  Have I told you how to test them using:
```
cd src/ingest-pipeline/misc/tools
./schema_tester.py export_and_backup_map_schema.yml ../../airflow/dags/export_and_backup_map.yml
```
Yours passes that test at the moment, but I think only because it's under-specified.

src/ingest-pipeline/airflow/dags/export_and_backup.py
```diff
@@ -0,0 +1,148 @@
...
+        info_dict = kwargs["ti"].xcom_pull(task_ids="find_plugins").copy()
+        for key in info_dict:
+            logging.info(f"{key.upper()}: {info_dict[key]}")
+        plugin_path = Path(export_and_backup_plugin.__file__).parent / "plugins"
+        print("plugin_path: ", plugin_path)
+        for plugin in export_and_backup_plugin.export_and_backup_result_iter(
+            plugin_path, **info_dict
+        ):
+
+            result = plugin.run_plugin()
```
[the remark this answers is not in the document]
jswelling: If you take this advice, all the individual plugins' run_plugin methods have to become .run_plugin(**info_dict) or something similar though.

jswelling (changes requested):
The schemata are getting tested correctly now, but the schema syntax needs a bit of work.  Getting close!

src/ingest-pipeline/airflow/dags/export_and_backup.py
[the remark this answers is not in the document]
jswelling: You're right and I was wrong about the plugin calling argument list.  My extraneous comment about .airflowignore still stands, though, so I'll create a new comment about that.

jswelling (changes requested):
Ooh, I remembered an important part- there is a hidden file in airflow/dags called .airflowignore which contains a list of things which Airflow should ignore and directories it should not enter. The directory export_and_backup and the map export_and_backup_map.yml need to be added to that file. Sorry; should have noticed that bit earlier.

src/ingest-pipeline/schemata/export_and_backup_map_schema.yml
gesinaphillips: I took a crack at this update. It still passes but I am uncertain whether to trust that or not now :)

src/ingest-pipeline/schemata/export_and_backup_map_schema.yml
gesinaphillips: Done!

src/ingest-pipeline/airflow/dags/.airflowignore
```diff
@@ -8,4 +8,4 @@ workflow_map.yml
 resource_map.yml
 celery_queue_map.yml
 diagnostics
-export_and_backup
+/export_and_backup
```
jswelling, line 11: I think the leading / before export_and_backup is the wrong syntax- 'diagnostics' is a directory and it seems to work without a /, and what little documentation I found for .airflowignore files doesn't suggest a / .  Do you think the / is needed?

jswelling (commented):
One last issue- maybe the '/' in the .airflowignore file is a stray character during editing?

src/ingest-pipeline/schemata/export_and_backup_map_schema.yml
jswelling: The schemata look right to me now.  If they pass the schema tester, I think you've got this part straightened out.

src/ingest-pipeline/airflow/dags/.airflowignore
jswelling: Ooh, good point. OK, that's a good reason for the '/'.

jswelling (approved):

Issue #693, opened 2022-12-14
fix instance/app.cfg.example


jswelling:
The [globus] block has moved from app.cfg to airflow.cfg .  Modify app.cfg.example to show this.

PR #697, opened 2022-12-15
The [globus] block moves from app.cfg to airflow.cfg

jswelling:
Remove the [globus] block from app.cfg.example, because it now goes in airflow.cfg .


jswelling:
closes https://github.com/hubmapconsortium/ingest-pipeline/issues/693

sunset666 (approved):

PR #699, opened 2023-01-03
Switch to Conda for all instances

sunset666:
This PR activates the base conda environment for all the instances of ingest-pipeline, which eases the process of upgrading Python on future releases


gesinaphillips (commented):
This seems pretty straightforward to me. I do not feel that I have sufficient expertise on conda environments to offer much critique, though.

PR #705, opened 2023-01-23
Bug fix:

DerekFurstPitt:
A line in a string has all of its hyphens replaced with commas for parsing, however later, a line is referenced that had not been similarly modified, causing an error. While the resulting output resembled the correct output, it was actually completely wrong. This has been fixed

Optimization:

Moved some areas around such that records associated with public endpoints are not inspected against audit logs. Since certain data is not collected from public datasets, it only wastes time to process them.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/generate_usage_report.py
+++ b/src/ingest-pipeline/airflow/dags/generate_usage_report.py
@@ -175,7 +175,7 @@ with HMDAG('generate_usage_report',
                 entity_type = "Public: Unknown"
             # for every line that contained the task id we are looking for, we now
             # look for the reference number
-            if is_public is not None:
+            if is_public is True:
                 for line in lines_with_tid:
                     log_entry = line.replace("-", "")
                     ref_index = log_entry.find("ref=") + 4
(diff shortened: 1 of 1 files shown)


src/ingest-pipeline/airflow/dags/generate_usage_report.py
```diff
@@ -156,40 +156,47 @@ def build_report(**kwargs):
...
-                    ref_index = b.find(reference_number)
-                    if ref_index == -1:
-                        matches = re.search(pattern, b)
-                        if matches is not None:
-                            ref_output = matches.group(1)
-                            break
-                if ref_output is not None:
-                    dataset_id = ref_output
-                    break
+            if is_public is not None:
```
sunset666, line 178: This validation will always be True since 'is_public' was just declared a couple of lines above with default value of False.

sunset666 (changes requested):

src/ingest-pipeline/airflow/dags/generate_usage_report.py
DerekFurstPitt: Good catch, you are correct. In an earlier version of this script, is_public was initialized as None rather than false. When stitching together my revisions into the existing version in ingest-pipeline, I mistakenly left this in. Thank you for finding that. This has been fixed and I've pushed the fix to `Derek-Furst/update-generate-usage-reports`.

sunset666 (approved):

PR #718, opened 2023-02-17
Welling/run background subtraction parallel control

jswelling:
**This branch cannot be deployed until a version of codex-pipeline downstream of both a4c9d72 and 4f78897 has been deployed**, since it invokes a command line argument added by those mods.  This branch adds a parameter to the invocation of codex-pipeline:run_background_subtraction.cwl in the codex_cytokit.py DAG which limits the amount of parallelism used by the Docker container in that step.  Previously parallelism was not specified so it defaulted to the number of cores on the node, which is not what we want.

This change does limit the parallelism in run_background_subtraction, which is the last unlimited bit of the DAG's invocation of the codex-pipeline repo.  It makes very little difference to running time in practice.  The parallelism of the SPRM repo is still unlimited, so the codex_cytokit DAG still presents parallelism problems.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/codex_cytokit.py
+++ b/src/ingest-pipeline/airflow/dags/codex_cytokit.py
@@ -29,6 +29,7 @@ from utils import (
     HMDAG,
     get_queue_resource,
     get_preserve_scratch_resource,
+    get_threads_resource,
 )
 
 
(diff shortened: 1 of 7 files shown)


jswelling:
Needs to be rebased.  Closed and recreated.

PR #720, opened 2023-02-17
Deltas found in place in the pittdev tree

jswelling:
These changes were found in place in the pittdev ingest-pipeline tree; they looked sensible but aren't in the current HEAD of devel.  I just grabbed them in case some of them were wanted.


sunset666:
This is part of the upgrade path to Python 3.9. So, not yet able to be added/merged into devel.

PR #724, opened 2023-03-03
Refactoring snd_status_msg function

sunset666:
Refactoring of the function used to send status update to entity-api to make use of the standar endpoint which allows to update more information.


src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -1238,6 +1238,218 @@ def send_status_msg(**kwargs) -> bool:
     return send_status_msg
 
 
+def make_send_status_msg_function(
+        dag_file: str,
+        retcode_ops: List[str],
+        cwl_workflows: List[Path],
+        http_conn_id: str = 'entity_api_connection',
```
jswelling, line 1245: It looks like it will cause a problem if http_conn_id is anything other than entity_api_connection now.  Maybe it's time for the whole argument to go away?

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -1238,6 +1238,218 @@ def send_status_msg(**kwargs) -> bool:
...
+        method = 'PUT'
+        headers = {
+            'authorization': 'Bearer ' + get_auth_tok(**kwargs),
+            'content-type': 'application/json',
+            'X-Hubmap-Application': 'ingest-pipeline',
+        }
+        extra_options = {}
+        return_status = True  # mark false on failure
+
+        http_hook = HttpHook(method, http_conn_id=http_conn_id)
```
jswelling, line 1334: Are there any places in the DAGs where http_conn_id is explicitly passed in?

jswelling (changes requested):
Is there any legit situation where http_conn_id would be passed in now?  In earlier days sometimes we would aim commits at different instances, so commits would happen on DEV in response to a run on PROD, but that never happens anymore.  I'm thinking that whole parameter should go, so that anything that tries to call with that parameter would generate an error.
The case where it's called from scan_and_begin_processing is strange, in that it sets the state to Invalid rather than Error- I think that's the only situation where that is expected to happen.  It's an awkward little kludge that I remember only vaguely.

src/ingest-pipeline/airflow/dags/utils.py
sunset666: looks like it is time. I'll remove it.

src/ingest-pipeline/airflow/dags/utils.py
sunset666: No, will remove that from the function since it is not used anymore.

PR #728, opened 2023-03-21
Bugfix send msg function + New RHEL 9 environment variables

sunset666:
- Bugfix on the new send msg function that updates entity metadata for different scenarios.
- New RHEL 9 environment scripts.


jpuerto-psc (approved):
New env files look good - can't speak to the bugfix.

PR #732, opened 2023-03-24
Pipeline error catching

gesinaphillips:
I'd love feedback on the base class (catch_pipeline_errors.py) and the subclass implementation in validate_upload.py.

Notes:
- The base FailureCallback class can be set as the callback method on a DAG to catch errors in the running DAG and send a notification to the internal email recipients set on the base class (likely data curators). It does not always need to be subclassed.
- However, in order to catch specific errors and send emails to data providers, FailureCallback must be subclassed.
- The ValidateUploadFailure subclass of FailureCallback is necessary in order to send emails when the validate_upload DAG succeeds but sub-processes determine the upload to be Invalid. Without this subclass, no notifications would be sent at all despite the upload entering an invalid state.
- Templates are examples only, the ones in validate_upload.py in particular are not good. Actual templates--in particular those to external users--would need more work.
- There's no reason this couldn't live in utils, I just broke it out into its own file to work on it.

If you want, you can test on dev using this branch. I've been using [this upload](https://ingest.dev.hubmapconsortium.org/upload/207ddcca543e91d6f18293804a89f38f) which has metadata errors. You can turn on one of the test exceptions in validate_upload > run_validation to see how it behaves if the DAG itself fails.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
+++ b/src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
@@ -16,100 +16,106 @@ TODO
 - type hinting
 """
 
+
 class FailureCallbackException(Exception):
     pass
 
 
 class FailureCallback:
+    """
+    List should be overwritten by each subclass with appropriate values.
+    """
+
+    external_exceptions = []
+    # TODO: Switch to curator email(s)
+    internal_email_recipients = ["gesina@psc.edu"]
+
+    def __init__(self, context, execute_methods=True):
+        self.context = context
+        self.uuid = self.context.get("task_instance").xcom_pull(key="uuid")
+        self.crypt_auth_tok = self.context.get("crypt_auth_tok")
+        self.dag_run = self.context.get("dag_run")
+        self.task = self.context.get("task")
+        self.exception = self.context.get("exception")
+        self.exception_name = type(self.exception).__name__
+
+        if execute_methods:
+            self.set_status()
+            self.send_failure_email()
+
+    # This is simplified from pythonop_get_dataset_state in utils
+    def get_submission_context(self):
+        method = "GET"
+        auth_tok = get_auth_tok(**{"crypt_auth_tok": self.crypt_auth_tok})
+        headers = {
+            "authorization": f"Bearer {auth_tok}",
+            "content-type": "application/json",
+            "X-Hubmap-Application": "ingest-pipeline",
+        }
+        http_hook = HttpHook(method, http_conn_id="ingest_api_connection")
+
+        endpoint = f"entities/{self.uuid}"
+
+        try:
+            response = http_hook.run(
+                endpoint, headers=headers, extra_options={"check_response": False}
+            )
+            response.raise_for_status()
+            submission_data = response.json()
+            return submission_data
+        except HTTPError as e:
+            print(f"ERROR: {e}")
+            if e.response.status_code == codes.unauthorized:
+                raise RuntimeError("entity database authorization was rejected?")
+            else:
+                print("benign error")
+                return {}
+
+    def get_status_and_message(self):
         """
-        List should be overwritten by each subclass with appropriate values.
+        Error message might need to be overwritten when
+        subclassed for various DAGs, so this is broken out.
         """
-        # Should probably be importing custom exceptions rather than comparing strings
-        external_exceptions = []
-        # TODO: Switch to curator email(s)
-        internal_email_recipients = ["gesina@psc.edu"]
+        return {
+            "status": "Invalid",
+            "validation_message": f"Process {self.dag_run.dag_id} started {self.dag_run.execution_date} failed at task {self.task.task_id} with error {self.exception_name} {self.exception}",
+        }
 
-        def __init__(self, context, execute_methods=True):
-            self.context = context
-            self.uuid = self.context.get("task_instance").xcom_pull(key="uuid")
-            self.crypt_auth_tok = self.context.get("crypt_auth_tok")
-            self.dag_run = self.context.get("dag_run")
-            self.task = self.context.get("task")
-            self.exception = self.context.get("exception")
-            self.exception_name = type(self.exception).__name__
+    def set_status(self):
+        """
+        The failure callback needs to set the
+        dataset status.
+        """
+        data = self.get_status_and_message()
+        auth_tok = get_auth_tok(**{"crypt_auth_tok": self.crypt_auth_tok})
+        endpoint = f"/entities/{self.uuid}"
+        headers = {
+            "authorization": "Bearer " + auth_tok,
+            "X-Hubmap-Application": "ingest-pipeline",
+            "content-type": "application/json",
+        }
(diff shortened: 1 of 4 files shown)


src/ingest-pipeline/airflow/dags/validate_upload.py
```diff
@@ -9,172 +9,245 @@
...
+)
 
-sys.path.append(airflow_conf.as_dict()['connections']['SRC_PATH']
-                .strip("'").strip('"'))
-from submodules import (ingest_validation_tools_upload,  # noqa E402
-                        ingest_validation_tools_error_report,
-                        ingest_validation_tests)
 sys.path.pop()
 
+
```
jpuerto-psc, line 40: Rather than importing the base FailureCallback class and having the ValidateUploadFailure class in here - consider moving the ValidateUploadFailure class into a separate file (maybe we have a directory where we define all of our failure classes). Reduces overall code in a singular file.

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
```diff
@@ -0,0 +1,163 @@
...
+        self.exception_name = type(self.exception).__name__
+
+        if execute_methods:
+            self.set_status()
+            self.send_failure_email()
+
+    # This is simplified from pythonop_get_dataset_state in utils
+    def get_submission_context(self):
+        method = "GET"
+        auth_tok = get_auth_tok(**{"crypt_auth_tok": self.crypt_auth_tok})
```
jpuerto-psc, line 49: Can probably move this to be in the constructor

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
```diff
@@ -0,0 +1,163 @@
...
+    def __init__(self, context, execute_methods=True):
+        self.context = context
+        self.uuid = self.context.get("task_instance").xcom_pull(key="uuid")
+        self.crypt_auth_tok = self.context.get("crypt_auth_tok")
+        self.dag_run = self.context.get("dag_run")
+        self.task = self.context.get("task")
+        self.exception = self.context.get("exception")
+        self.exception_name = type(self.exception).__name__
+
+        if execute_methods:
```
jpuerto-psc, line 42: This is fine for a first iteration as we've discussed. We will want to think through how we can make the functions that are called (set_status, send_failure_email) a bit more dynamic though as we add additional "reporting" methods. IE, updating Asana, Slack etc.

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
```diff
@@ -0,0 +1,163 @@
...
+class FailureCallbackException(Exception):
+    pass
+
+
+class FailureCallback:
+    """
+    List should be overwritten by each subclass with appropriate values.
+    """
+
+    external_exceptions = []
```
jpuerto-psc, line 29: Fine for first iteration.

Might want to think about making this more of a mapping of exceptions to a list of "reporting" functions

src/ingest-pipeline/airflow/dags/validate_upload.py
gesinaphillips: Makes sense to me. I think a directory is probably the way to go and will create one.

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
gesinaphillips: Agreed, but I will do some thinking!

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
gesinaphillips: Done!

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
gesinaphillips: Made an update here, but I'm not really sure if I'm moving in the right direction.

src/ingest-pipeline/airflow/dags/catch_pipeline_errors.py
gesinaphillips: Do you mean like a list of notifications that should  be sent per specific error type, in the case where we know certain errors should be handled in specific ways? e.g.:
PreflightError: [send_failure_email, send_asana_notification],
ValidatorError: [send_failure_email]

My thought here is that perhaps it would make sense to separate internal and external email notifications into two different methods. Haven't totally thought that through yet. At the moment all errors generate an internal email at a minimum. That should be easy enough to keep doing unless no email is a better default.

jpuerto-psc (approved):

src/ingest-pipeline/airflow/dags/error_catching/failure_callback.py
```diff
@@ -0,0 +1,171 @@
...
+
+    # This is simplified from pythonop_get_dataset_state in utils
+    def get_submission_context(self):
+        method = "GET"
+        headers = {
+            "authorization": f"Bearer {self.auth_tok}",
+            "content-type": "application/json",
+            "X-Hubmap-Application": "ingest-pipeline",
+        }
+        http_hook = HttpHook(method, http_conn_id="ingest_api_connection")
```
sunset666, line 61: Maybe using entity_api_connection would be better, since we are moving toward the standardized endpoints suggested by Pitt

src/ingest-pipeline/airflow/dags/validate_upload.py
```diff
@@ -15,166 +15,188 @@
...
+        kwargs["ti"].xcom_push(key="validation_file_path", value=str(validation_file_path))
 
     t_run_validation = PythonOperator(
-        task_id='run_validation',
+        task_id="run_validation",
         python_callable=run_validation,
         provide_context=True,
         op_kwargs={
-        }
+            "crypt_auth_tok": (
```
sunset666, line 146: Is this really necessary? the token is already there in the DAG, you can use it as it is used in send_status_msg

src/ingest-pipeline/airflow/dags/error_catching/failure_callback.py
gesinaphillips: Done!

src/ingest-pipeline/airflow/dags/validate_upload.py
gesinaphillips: I struggled a little bit with getting the crypt_auth_tok to the failure callback. The failure callback receives context as its only argument, which does not include the dag's kwargs. I could xcom_push crypt_auth_tok but that would mean that, if the dag failed prior to that point, it would not have the token. I could do an xcom_push in the first line of the first function (find_uuid) instead of requiring op_kwargs to be passed to every task; this would have the benefit of being a single line fix, and should limit the cases where the dag can fail without first passing the necessary token to the failure callback. Of course, I could be missing something here! Please let me know what you think.

PR #734, opened 2023-03-28
New DAG to rebuild metadata

sunset666:
This new DAG will re-process metadata for a uuid to be updated for missing/changed information.


src/ingest-pipeline/airflow/dags/rebuild_metadata.py
```diff
@@ -0,0 +1,198 @@
...
+    'owner': 'hubmap',
+    'depends_on_past': False,
+    'start_date': datetime(2019, 1, 1),
+    'email': ['joel.welling@gmail.com'],
+    'email_on_failure': False,
+    'email_on_retry': False,
+    'retries': 1,
+    'retry_delay': timedelta(minutes=1),
+    'xcom_push': True,
+    'queue': get_queue_resource('launch_multi_analysis'),
```
jswelling, line 56: Did you mean to leave this pointed at launch_multi_analysis?  Probably OK but I wanted to make sure it's intentional, since it violates the usual pattern.

jswelling (changes requested):

src/ingest-pipeline/airflow/dags/rebuild_metadata.py
sunset666: Good catch, I've updated to reflect the correct DAG.

jswelling (approved):

PR #742, opened 2023-04-07
Export dataset information upon publication based on Ivan's hubmap-inventory code

gesinaphillips:
The dataset_published plugin for the export_and_backup DAG is designed to be integrated into the workflow directly after a dataset's status is changed to "Published." This plugin calls the `create` function in Ivan's hubmap-inventory submodule in order to create the files he needs for CFDE submissions.

This has been tested on vm001 with Ivan and it works. It writes to /hive/hubmap/bdbags/inventory. The `backup` argument passed to the `create` function is currently set to False to avoid writing data from any environment other than prod. On prod, that parameter should be set to True.


.gitmodules
```diff
@@ -52,3 +52,9 @@
 [submodule "src/ingest-pipeline/submodules/py-hubmapbags"]
 	path = src/ingest-pipeline/submodules/py-hubmapbags
 	url = https://github.com/hubmapconsortium/py-hubmapbags.git
+[submodule "src/ingest-pipeline/airflow/dags/hubmap-inventory"]
+	path = src/ingest-pipeline/airflow/dags/hubmap-inventory
+	url = https://github.com/hubmapconsortium/hubmap-inventory
+[submodule "src/ingest-pipeline/submodules/hubmap-inventory"]
```
sunset666, line 58: Does it need to be in two different places?

src/ingest-pipeline/airflow/dags/export_and_backup/plugins/dataset_published.py
```diff
@@ -1,8 +1,81 @@
...
+
+    # Should I add this to utils instead, so it can be reused?
+    def get_hubmap_id(self, uuid):
+        method = 'GET'
+        headers = {
+            'authorization': f'Bearer {self.token}',
+            'content-type': 'application/json',
+            'X-Hubmap-Application': 'ingest-pipeline',
+            }
+        http_hook = HttpHook(method, http_conn_id='ingest_api_connection')
```
sunset666, line 43: Let's use entity_api_connection to query entities

.gitmodules
gesinaphillips: Whhoooops

src/ingest-pipeline/airflow/dags/export_and_backup/plugins/dataset_published.py
gesinaphillips: Updated!

PR #755, opened 2023-05-30
Draft status change and Asana updater logic

gesinaphillips:
This PR encompasses logic to package a few things together on status change:
- If changing to Error, use FailureCallback first
- Change status in Entity API
- Update Asana card status and/or note in workflow management system

**FailureCallback** has been refactored to be simpler, removing the status change and notification logic from it.
**status_manager** includes both StatusChanger and UpdateAsana
- **StatusChanger** handles making the call to Entity API to change status and triggering Asana update; can be extended to do additional things if needed (e.g. send emails to submitters)
- **UpdateAsana** interfaces with the workflow management system Asana board. Maps new upload/dataset status to Asana status and changes accordingly.
> - There are unsolved problems here regarding uploads vs. datasets. Uploads are (I think) handled. Datasets are more troublesome. There are two ways that datasets are handled in Asana:
>> - Datasets that are part of uploads are not listed in the Asana task, but all fall under the parent upload HuBMAP ID.
>> - Datasets are uploaded individually, and may be bundled together into a task but do not have any other encoded relation to one another.
> - Additional complexity: all datasets in an upload or dataset group must be in the right status (e.g. QA, Approved, Published) for the card to move forward. I.e. if one dataset reaches QA but all others are in the Error state, we do not want to move the Asana task to Post-Processing! We want it to move to or stay in the Asana Blocked status.
> - This is where I have been focusing most recently.

Additional notes:
- Most changes to **utils.py** are aesthetic due to linting, except for pythonop_set_dataset_state, which has been refactored to use new infrastructure.
- Please ignore **gen_pub_ancillary.py**, it crept in there during a rebase, oops.
- Currently working on testing in Airflow. Have tested basic path for Invalid/Error status uploads, now honestly having some trouble creating a valid upload. Dataset paths need a bit more testing.


gesinaphillips:
Added @jpuerto-psc and @sunset666 in advance of code review meeting.

PR #757, opened 2023-06-05
Bugfix Publication survey

sunset666:
Adding missing Publication assignment when building the Entity object.


jswelling (commented):
Oopsie; I guess I missed that one :-/

PR #774, opened 2023-08-14
Submitted status for split datasets from Upload

sunset666:
New datasets created on reorganize DAG will get the Submitted status since they are already validated.


bhonick:
This will be helpful, thanks!

gesinaphillips (approved):
I'll want to switch the call to entity-api over to the unified StatusChanger that I've designed but that's been hung up with the Asana work so we can put this out if it's meant to get deployed soon!

PR #782, opened 2023-10-09
StatusChanger

gesinaphillips:
Standardizes how statuses are changed in ingest-pipeline. This is a prerequisite for Asana notifications (unless we want to switch over to subscribing to/listening an endpoint and picking up status changes from there). Either way, I think the standardization this introduces is potentially nicer and more extensible than writing the entity-api call from scratch every time.

I spent some time retooling how I was planning to do email notifications. Those are basically done if we decide to go back to that idea but at this point they were a distraction from the main goal of getting this basic piece of infrastructure in place.


src/ingest-pipeline/airflow/dags/status_change/status_manager.py
```diff
@@ -0,0 +1,239 @@
...
+    UPLOAD_INVALID = "invalid"
+    UPLOAD_NEW = "new"
+    UPLOAD_PROCESSING = "processing"
+    UPLOAD_REORGANIZED = "reorganized"
+    UPLOAD_SUBMITTED = "submitted"
+    UPLOAD_VALID = "valid"
+
+
+# Needed some way to disambiguate statuses shared by datasets and uploads
+ENTITY_STATUS_MAP = {
```
sunset666, line 34: Probably needs Publication to be added to the map.

PR #790, opened 2023-11-20
New DAG for annotations

sunset666:
This new DAG will create a new versioning for a previous uuid running just azimuth annotations. It copies over files from the previous run into this new dataset to be processed with the downstream pipelines for annotations.


sunset666:
@gesinaphillips, @jswelling. I didn't find an elegant way to get the correct pipeline it was run on the previous version and the new Azimuth pipeline requires a new parameter that is directly dependent on that, so it gets into this multiple if - else function to map it.

gesinaphillips (approved):

PR #799, opened 2023-12-05
Nihdev 441 pipeline mapping changes for soft assay

jpuerto-psc:
- Updates the metadata extraction so that it works specifically for the parent of a multi-assay group. Sunset is working on getting that to work for the components as well.
- This PR also updates the pipeline decider such that it grabs the soft assay type for the dataset and passes that through to the utils.downstream_workflow_iter
- This PR also points us to latest IVT which contains the validation changes necessary for multi-assay validation. This can be broken out and be released separately if we get movement on the ingest-api front.


src/ingest-pipeline/md/data_collection_types/multiassay_metadatatsv_data_collection.py
```diff
@@ -0,0 +1,143 @@
...
+
+import requests
+from requests import codes
+from requests.exceptions import HTTPError
+from type_base import MetadataError
+from data_collection import DataCollection
+
+
+class MultiassayMetadataTSVDataCollection(DataCollection):
+    category_name = 'MULTIASSAYMETADATATSV';
```
sunset666, line 20: trailing semicolon

src/ingest-pipeline/md/data_collection_types/multiassay_metadatatsv_data_collection.py
```diff
@@ -0,0 +1,143 @@
...
+                        else:
+                            print('benign error')
+                            return None
+
+                    if 'metadata' in fname and fname.endswith('.tsv'):
+                        assert isinstance(this_md, list), 'metadata.tsv did not produce a list'
+                        if 'must-contain' in response:
+                            print('MULTI ASSAY FOUND')
+                            for rec in this_md:
+                                for key in ['dataset_type', 'data_path', 'contributors_path']:
```
sunset666, line 99: Is this needed? this step happens after the metadata.tsvs were already validated

src/ingest-pipeline/md/data_collection_types/multiassay_metadatatsv_data_collection.py
jpuerto-psc: Yeah, this was on the other datacollection class, I'll remove it

src/ingest-pipeline/md/data_collection_types/multiassay_metadatatsv_data_collection.py
jpuerto-psc: Are you referring to the for loop that checks the keys? I can remove it, was just copy-pasta from the other data collection class

sunset666: actually the validation of required columns in the metadata file

PR #801, opened 2023-12-07
Automated previous version uuid

sunset666:
Previous version uuid is automatically added if there is a derived dataset in Publish state, avoiding the need of manually finding this and simplifying re-processing.


src/ingest-pipeline/airflow/dags/launch_multi_analysis.py
```diff
@@ -17,12 +17,14 @@
...
 
+from extra_utils import check_link_published_drvs
+
 
 def get_uuid_for_error(**kwargs):
     """
     Return the uuid for the derived dataset if it exists, and of the parent dataset otherwise.
     """
-    return None
+    return ''
```
jpuerto-psc, line 27: Can we get rid of this function and use a lambda instead in 'on_failure_callback'?

jpuerto-psc (approved):

PR #810, opened 2024-01-03
Reorganize refactor

sunset666:
Refactoring Reorganize Upload to support Soft / Multi Assay.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/extra_utils.py
+++ b/src/ingest-pipeline/airflow/dags/extra_utils.py
@@ -42,6 +42,7 @@ class SoftAssayClient:
     def __init__(self, metadata_files: List, auth_tok: str):
         self.assay_components = []
         self.primary_assay = {}
+        self.is_multiassay = True
         for metadata_file in metadata_files:
             try:
                 rows = self.__read_rows(metadata_file, encoding="UTF-8")
@@ -69,6 +70,7 @@ class SoftAssayClient:
                 print(f"Primary {self.primary_assay}")
         if not self.primary_assay and len(self.assay_components) == 1:
             self.primary_assay = self.assay_components.pop()
+            self.is_multiassay = False
 
     def __get_assaytype_data(
         self,
--- a/src/ingest-pipeline/airflow/dags/extra_utils.py
+++ b/src/ingest-pipeline/airflow/dags/extra_utils.py
@@ -1,30 +1,116 @@
-from pprint import pprint
-from typing import Tuple
+import json
+from typing import List, Dict
+from pathlib import Path
+from csv import DictReader
 
 from airflow.providers.http.hooks.http import HttpHook
 from airflow.configuration import conf as airflow_conf
+from pprint import pprint
+from typing import Tuple
 
 
 def check_link_published_drvs(uuid: str) -> Tuple[bool, str]:
     needs_previous_version = False
-    published_uuid = ''
-    endpoint = f'/children/{uuid}'
-    auth_tok = ''.join(e for e in airflow_conf.as_dict()['connections']['APP_CLIENT_SECRET'] if e.isalnum())
+    published_uuid = ""
+    endpoint = f"/children/{uuid}"
+    auth_tok = "".join(
+        e for e in airflow_conf.as_dict()["connections"]["APP_CLIENT_SECRET"] if e.isalnum()
+    )
     headers = {
-        'authorization': 'Bearer ' + auth_tok,
-        'content-type': 'application/json',
-        'X-Hubmap-Application': 'ingest-pipeline'}
+        "authorization": "Bearer " + auth_tok,
+        "content-type": "application/json",
+        "X-Hubmap-Application": "ingest-pipeline",
+    }
     extra_options = {}
 
-    http_hook = HttpHook('GET', http_conn_id='entity_api_connection')
+    http_hook = HttpHook("GET", http_conn_id="entity_api_connection")
 
-    response = http_hook.run(endpoint,
-                             headers=headers,
-                             extra_options=extra_options)
-    print('response: ')
+    response = http_hook.run(endpoint, headers=headers, extra_options=extra_options)
+    print("response: ")
     pprint(response.json())
     for data in response.json():
-        if data.get('entity_type') in ('Dataset', 'Publication') and data.get('status') == 'Published':
+        if (
+            data.get("entity_type") in ("Dataset", "Publication")
+            and data.get("status") == "Published"
+        ):
             needs_previous_version = True
-            published_uuid = data.get('uuid')
+            published_uuid = data.get("uuid")
     return needs_previous_version, published_uuid
+
+
+class SoftAssayClient:
+    def __init__(self, metadata_files: List, auth_tok: str):
+        self.assay_components = []
+        self.primary_assay = {}
+        for metadata_file in metadata_files:
+            try:
+                rows = self.__read_rows(metadata_file, encoding="UTF-8")
+            except Exception as e:
+                print(f"Error {e} reading metadata {metadata_file}")
+                return
+            assay_type = self.__get_assaytype_data(row=rows[0], auth_tok=auth_tok)
+            if not assay_type.get("must-contain"):
+                print(f"Component {assay_type}")
+                data_component = {
+                    "assaytype": assay_type.get("assaytype"),
+                    "contains-pii": assay_type.get("contains-pii", True),
+                    "primary": assay_type.get("primary", False),
+                    "metadata-file": metadata_file,
+                }
+                self.assay_components.append(data_component)
+            else:
+                data_component = {
+                    "assaytype": assay_type.get("assaytype"),
+                    "contains-pii": assay_type.get("contains-pii", True),
+                    "primary": assay_type.get("primary", False),
+                    "metadata-file": metadata_file,
+                }
+                self.primary_assay = data_component
+                print(f"Primary {self.primary_assay}")
+        if not self.primary_assay and len(self.assay_components) == 1:
+            self.primary_assay = self.assay_components.pop()
+
+    def __get_assaytype_data(
+        self,
+        row: Dict,
+        auth_tok: str,
+    ) -> Dict:
+        http_hook = HttpHook("POST", http_conn_id="ingest_api_connection")
+        endpoint = f"/assaytype"
+        headers = {
+            "Authorization": "Bearer " + auth_tok,
+            "Content-Type": "application/json",
+        }
+        response = http_hook.run(endpoint=endpoint, headers=headers, data=json.dumps(row))
+        response.raise_for_status()
+        return response.json()
+
+    def __get_context_of_decode_error(self, e: UnicodeDecodeError) -> str:
+        buffer = 20
+        codec = "latin-1"  # This is not the actual codec of the string!
+        before = e.object[max(e.start - buffer, 0) : max(e.start, 0)].decode(codec)  # noqa
+        problem = e.object[e.start : e.end].decode(codec)  # noqa
+        after = e.object[e.end : min(e.end + buffer, len(e.object))].decode(codec)  # noqa
+        in_context = f"{before} [ {problem} ] {after}"
+        return f'Invalid {e.encoding} because {e.reason}: "{in_context}"'
+
+    def __dict_reader_wrapper(self, path, encoding: str) -> list:
+        with open(path, encoding=encoding) as f:
+            rows = list(DictReader(f, dialect="excel-tab"))
+        return rows
+
+    def __read_rows(self, path: Path, encoding: str) -> List:
+        if not Path(path).exists():
+            message = {"File does not exist": f"{path}"}
+            raise message
+        try:
+            rows = self.__dict_reader_wrapper(path, encoding)
+            if not rows:
+                message = {"File has no data rows": f"{path}"}
+            else:
+                return rows
+        except IsADirectoryError:
+            message = {"Expected a TSV, but found a directory": f"{path}"}
+        except UnicodeDecodeError as e:
+            message = {"Decode Error": self.__get_context_of_decode_error(e)}
+        raise message
--- a/src/ingest-pipeline/airflow/dags/extra_utils.py
+++ b/src/ingest-pipeline/airflow/dags/extra_utils.py
@@ -4,7 +4,6 @@ from pathlib import Path
 from csv import DictReader
 
 from airflow.providers.http.hooks.http import HttpHook
-from airflow.configuration import conf as airflow_conf
 from pprint import pprint
 from typing import Tuple
 
@@ -13,11 +12,7 @@ def check_link_published_drvs(uuid: str) -> Tuple[bool, str]:
     needs_previous_version = False
     published_uuid = ""
     endpoint = f"/children/{uuid}"
-    auth_tok = "".join(
-        e for e in airflow_conf.as_dict()["connections"]["APP_CLIENT_SECRET"] if e.isalnum()
-    )
     headers = {
-        "authorization": "Bearer " + auth_tok,
         "content-type": "application/json",
         "X-Hubmap-Application": "ingest-pipeline",
     }
--- a/src/ingest-pipeline/airflow/dags/extra_utils.py
+++ b/src/ingest-pipeline/airflow/dags/extra_utils.py
@@ -50,24 +50,18 @@ class SoftAssayClient:
                 print(f"Error {e} reading metadata {metadata_file}")
                 return
             assay_type = self.__get_assaytype_data(row=rows[0], auth_tok=auth_tok)
+            data_component = {
+                "assaytype": assay_type.get("assaytype"),
+                "contains-pii": assay_type.get("contains-pii", True),
+                "primary": assay_type.get("primary", False),
+                "metadata-file": metadata_file,
+            }
             if not assay_type.get("must-contain"):
                 print(f"Component {assay_type}")
-                data_component = {
-                    "assaytype": assay_type.get("assaytype"),
-                    "contains-pii": assay_type.get("contains-pii", True),
-                    "primary": assay_type.get("primary", False),
-                    "metadata-file": metadata_file,
-                }
                 self.assay_components.append(data_component)
             else:
-                data_component = {
-                    "assaytype": assay_type.get("assaytype"),
-                    "contains-pii": assay_type.get("contains-pii", True),
-                    "primary": assay_type.get("primary", False),
-                    "metadata-file": metadata_file,
-                }
+                print(f"Primary {assay_type}")
                 self.primary_assay = data_component
-                print(f"Primary {self.primary_assay}")
         if not self.primary_assay and len(self.assay_components) == 1:
             self.primary_assay = self.assay_components.pop()
             self.is_multiassay = False
--- a/src/ingest-pipeline/airflow/dags/reorganize_upload.py
+++ b/src/ingest-pipeline/airflow/dags/reorganize_upload.py
@@ -104,31 +104,6 @@ with HMDAG(
             scanned_md = yaml.safe_load(f)
         return scanned_md
 
-    def __get_lzpath_uuid(**kwargs):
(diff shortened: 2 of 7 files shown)


src/ingest-pipeline/misc/tools/split_and_create.py
```diff
@@ -123,13 +130,7 @@ def create_new_uuid(row, source_entity, entity_factory, dryrun=False):
...
     assert info_txt_root is not None, "Expected a Dataset or an Upload"
     info_txt = info_txt_root + " : " + rec_identifier
-    try:
-        type_info = entity_factory.type_client.getAssayType(canonical_assay_type)
-    except Exception:
-        print(f"tried {orig_assay_type}, canoncal version {canonical_assay_type}")
-        print(f"options are {list(entity_factory.type_client.iterAssayNames())}")
-        type_info = entity_factory.type_client.getAssayType(orig_assay_type)
-    contains_human_genetic_sequences = type_info.contains_pii
+    contains_human_genetic_sequences = primary_entity.get("contains-pii")
```
jpuerto-psc, line 133: Is this taking care of the refactor that Zhou mentioned to get rid of the type_client?

src/ingest-pipeline/misc/tools/split_and_create.py
```diff
@@ -338,31 +369,36 @@ def reorganize(source_uuid, **kwargs) -> None:
...
     entity_factory = EntityFactory(auth_tok, instance=instance)
 
     print(f"Decomposing {source_uuid}")
     source_entity = entity_factory.get(source_uuid)
+    full_entity = SoftAssayClient(
+        list(source_entity.full_path.glob("*metadata.tsv")), instance, auth_tok
+    )
     if mode in ["all", "stop"]:
-        if hasattr(source_entity, "data_types"):
+        if hasattr(source_entity, "dataset_type"):
```
jpuerto-psc, line 383: I think this should be fine, uploads should have a dataset_type specified.

src/ingest-pipeline/misc/tools/survey.py
```diff
@@ -702,7 +702,7 @@ def create_dataset(self,
         ingest_url = ENDPOINTS[self.instance]['ingest_url']
         data = {"provider_info": provider_info,
                 "contains_human_genetic_sequences": contains_human_genetic_sequences,
-                "data_types": [assay_type],
```
jpuerto-psc, line 705: I'm a bit concerned here - I believe that Joel's code relies on the data_types field for the old style data, so if we were to process any old style data/create datasets, this will break the assay classifier since the data_types would be blank.

jpuerto-psc: @jswelling could you confirm whether my assumption here is valid or not?

jpuerto-psc (changes requested):

src/ingest-pipeline/misc/tools/split_and_create.py
sunset666: Yes, it probably needs more work around that in other parts of the code for different functionality.

src/ingest-pipeline/misc/tools/split_and_create.py
sunset666: They should, that's generated in the ingest-ui (for uploads)

src/ingest-pipeline/airflow/dags/extra_utils.py
```diff
@@ -1,30 +1,118 @@
...
+        for metadata_file in metadata_files:
+            try:
+                rows = self.__read_rows(metadata_file, encoding="UTF-8")
+            except Exception as e:
+                print(f"Error {e} reading metadata {metadata_file}")
+                return
+            assay_type = self.__get_assaytype_data(row=rows[0], auth_tok=auth_tok)
+            if not assay_type.get("must-contain"):
+                print(f"Component {assay_type}")
+                data_component = {
```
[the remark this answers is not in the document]
sunset666: Good catch.

src/ingest-pipeline/airflow/dags/extra_utils.py
```diff
@@ -1,30 +1,112 @@
...
+
+    def __dict_reader_wrapper(self, path, encoding: str) -> list:
+        with open(path, encoding=encoding) as f:
+            rows = list(DictReader(f, dialect="excel-tab"))
+        return rows
+
+    def __read_rows(self, path: Path, encoding: str) -> List:
+        if not Path(path).exists():
+            message = {"File does not exist": f"{path}"}
+            raise message
```
gesinaphillips, line 101: Are you intentionally not raising an exception here and on line 112?

src/ingest-pipeline/misc/tools/split_and_create.py
```diff
@@ -338,31 +369,36 @@ def reorganize(source_uuid, **kwargs) -> None:
...
+            # ToDo: review this change to dataset_type
             assert isinstance(source_entity.data_types, str)
             source_data_types = source_entity.data_types
         else:
             source_data_types = None
-        source_metadata_files = list(source_entity.full_path.glob("*metadata.tsv"))
-        for src_idx, smf in enumerate(source_metadata_files):
+        for src_idx, smf in enumerate(full_entity.primary_assay.get("metadata-file")):
```
gesinaphillips, line 389: Do we anticipate that there will be multiple metadata files in full_entity.primary_assay["metadata-file"]? SoftAssayClient doesn't seem to allow for that. Trying to figure out if this enumeration is still necessary here and on 415 (but admittedly this code is a little tangled for me).

src/ingest-pipeline/misc/tools/split_and_create.py
```diff
@@ -199,8 +200,38 @@ def populate(row, source_entity, entity_factory, dryrun=False):
...
         kid_path = Path(entity_factory.get_full_path(uuid))
+
     row_df.to_csv(kid_path / f"{uuid}-metadata.tsv", header=True, sep="\t", index=False)
     extras_path = kid_path / "extras"
+
+    # ToDo read all component metadata files find the matching row for old_data_path and then refactor info
+    #  as main metadata above before creating the new component metadata tsv
+    if components is not None:
+        for component in components:
+            component_df = pd.read_csv(component.get("metadata-file"))
```
gesinaphillips, line 211: I think that the component metadata file is still a TSV by this point. You might need `sep="\t"` here.

src/ingest-pipeline/misc/tools/split_and_create.py
```diff
@@ -199,8 +200,38 @@ def populate(row, source_entity, entity_factory, dryrun=False):
...
+            row_component = component_df.query(f'data_path=="{old_data_path}"')
+            assert (
+                len(row_component) == 0
```
gesinaphillips, line 214: I see your TODO at the top of this function so maybe this is not completely ready yet, but I'm a little confused here. On line 211 you read in the component metadata file and then on line 212 you query it for a `data_path` field matching the `old_data_path` found in the primary TSV. I would be expecting you to want this to find a match (indicating that the component TSV has a row matching the row we're looking at in the parent TSV) and for this assert to read `len(row_components) == 1` rather than 0. Am I just confused?

src/ingest-pipeline/misc/tools/survey.py
jswelling: @jpuerto-psc @sunset666 yeah, I'm concerned about the logic at this point.  CEDAR-style input records will have a less-specific 'dataset_type' field rather than the completely specific 'assay_type' field, and the call from split_and_create to create_dataset doesn't seem to allow for this new structure.  I suspect create_dataset has to accept either an assay_type or a dataset_type, and needs to build the metadata record for the new dataset appropriately in either case.  Those records would look like old- and new- style dataset metadata.  If only new-style is meant to be supported, this code won't do it.

src/ingest-pipeline/airflow/dags/reorganize_upload.py
```diff
@@ -19,208 +24,398 @@
...
+    },
+) as dag:
+
+    def read_metadata_file(uuid, **kwargs):
+        md_fname = os.path.join(utils.get_tmp_dir_path(kwargs["run_id"]), uuid + "-" + "rslt.yml")
+        with open(md_fname, "r") as f:
+            scanned_md = yaml.safe_load(f)
+        return scanned_md
+
+    def __get_lzpath_uuid(**kwargs):
```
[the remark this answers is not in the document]
sunset666: Great catch, bad copy and paste.

src/ingest-pipeline/airflow/dags/extra_utils.py
sunset666: I'm, this will be like a 'not found' response but in the form an exception (copied from IVT, basically). If the line is correct we will have row returned.

src/ingest-pipeline/misc/tools/split_and_create.py
sunset666: Great catch

src/ingest-pipeline/misc/tools/split_and_create.py
sunset666: You are right, it should be 1, somehow I had it backwards. Removing the TODO.

src/ingest-pipeline/misc/tools/split_and_create.py
sunset666: We probably won't, this was a straight forward replacement of the file reading with the new multiassay primary. It doesn't break anything leaving it as is (it will just have 0 as index). I agree, that part of the code is quite tangled, maybe we can ask @jswelling what the intention was of having that index in place.

src/ingest-pipeline/airflow/dags/extra_utils.py
```diff
@@ -1,30 +1,112 @@
...
+                print(f"Component {assay_type}")
+                self.assay_components.append(data_component)
+            else:
+                print(f"Primary {assay_type}")
+                self.primary_assay = data_component
+        if not self.primary_assay and len(self.assay_components) == 1:
+            self.primary_assay = self.assay_components.pop()
+            self.is_multiassay = False
+
+    def __get_assaytype_data(
```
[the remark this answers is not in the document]
sunset666: it doesn't hurt having it, but maybe we can remove it. Let's wait for @jpuerto-psc.

src/ingest-pipeline/misc/tools/split_and_create.py
gesinaphillips: Also agreed that this is probably just an aesthetic / clarity concern rather than a functional one.

src/ingest-pipeline/airflow/dags/extra_utils.py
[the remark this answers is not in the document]
gesinaphillips: Agreed, no harm in it being there, have just run into enough auth errors that I love a chance to delete code requiring a globus token :P

src/ingest-pipeline/misc/tools/survey.py
sunset666: @jswelling, @jpuerto-psc, we might need to setup a discussion around this since the approach here was to use the new softassay classifier to get the assay_type.

gesinaphillips:
IVT will validate a multi-assay upload where one or more children also point to standalone datasets. That is, say you have the following upload:

```
visium-metadata.tsv
histology-metadata.tsv
rnaseq-metadata.tsv
dataset-1/
dataset-2/
```

All three metadata TSVs include `dataset-1` as a data_path. However, only histology-metadata.tsv includes `dataset-2` as a data_path, because `dataset-2` is not part of the multi-assay upload. As far as IVT is concerned, this is valid. Would `dataset-2` be reorganized properly when it got to that stage? Having a little trouble following this through on my own.

Tagging @jpuerto-psc here too because he was involved in this discussion about standalone datasets inside of a multi-assay upload.

src/ingest-pipeline/misc/tools/survey.py
gesinaphillips: Please loop me in too @sunset666 :)

sunset666:
> IVT will validate a multi-assay upload where one or more children also point to standalone datasets. That is, say you have the following upload:
> 
> ```
> visium-metadata.tsv
> histology-metadata.tsv
> rnaseq-metadata.tsv
> dataset-1/
> dataset-2/
> ```
> 
> All three metadata TSVs include `dataset-1` as a data_path. However, only histology-metadata.tsv includes `dataset-2` as a data_path, because `dataset-2` is not part of the multi-assay upload. As far as IVT is concerned, this is valid. Would `dataset-2` be reorganized properly when it got to that stage? Having a little trouble following this through on my own.
> 
> Tagging @jpuerto-psc here too because he was involved in this discussion about standalone datasets inside of a multi-assay upload.

Since this is just going to work on Uploads, I'm expecting it won't affect the reorganization process (because it will mainly create the datasets specified in the big files, which all will be pointing to the primary dataset folder on that upload). And the split of components don't really move anything, they just get create the 'virtual' dataset and the data path will point to the primary.

src/ingest-pipeline/airflow/dags/reorganize_upload.py
```diff
@@ -19,208 +24,373 @@
...
+        trigger_rule="all_done",
+    )
+
+    t_log_info = LogInfoOperator(task_id="log_info")
+
+    t_join = JoinOperator(task_id="join")
+
+    def flex_maybe_multiassay_spawn(**kwargs):
+        """
+        This is a generator which returns appropriate DagRunOrders
```
jswelling, line 336: This doesn't appear to be a generator any more, and it doesn't return anything, so the docstring should be fixed.

src/ingest-pipeline/misc/tools/split_and_create.py
jswelling: It used to be legit to have multiple (non-multiome) metadata.tsv files in the same upload directory, so that enumeration was meant to cope with them all in order.  Each would control distinct data subdirectories so there was basically no conflict.  I think we've deprecated the notion of multiple metadata.tsv files in cases which are not multiome, though.

src/ingest-pipeline/misc/tools/survey.py
jswelling: Yeah- I'm honestly having a hard time visualizing the logic flow.

jswelling (dismissed):

src/ingest-pipeline/airflow/dags/extra_utils.py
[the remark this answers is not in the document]
jpuerto-psc: Let's remove it

src/ingest-pipeline/misc/tools/split_and_create.py
jpuerto-psc: I don't think that the validation routines will report an error if there are multiple metadata.tsv files for non-multiome uploads. @gesinaphillips - is that the case?

src/ingest-pipeline/misc/tools/survey.py
jpuerto-psc: @jswelling - are you avail for our IEC dev meeting tomorrow? I'd like to discuss this then.

src/ingest-pipeline/airflow/dags/reorganize_upload.py
sunset666: Updated with the appropriate description

src/ingest-pipeline/misc/tools/split_and_create.py
jpuerto-psc: I take it back, I actually don't think that this is possible for non-multiome uploads as none of our metadata extraction classes allow for more than a single metadata.tsv file to be present in the root of an upload/dataset.

jpuerto-psc:
> Since this is just going to work on Uploads, I'm expecting it won't affect the reorganization process (because it will mainly create the datasets specified in the big files, which all will be pointing to the primary dataset folder on that upload). And the split of components don't really move anything, they just get create the 'virtual' dataset and the data path will point to the primary.

But are we detecting that the second Histology dataset is not technically part of a Multiome dataset? In Gesina's example, we would have two lines in the Histology TSV, and one line each in Visium and RNAseq TSVs. That second line in the Histology TSV which points to `dataset-2` should be treated as a non-multiome dataset. I don't think that this scenario is currently supported by the metadata extraction classes.

src/ingest-pipeline/misc/tools/split_and_create.py
gesinaphillips: A mixed upload will pass validation in most cases. However, this is currently not the case for a multi-assay upload, which will break if you include *-metadata.tsv files for any other dataset_types EXCEPT its children, which can refer to datasets that are not part of the parent upload (sorry, confusing).

So while this Visium example in my comment below is valid:

```
visium-metadata.tsv
histology-metadata.tsv
rnaseq-metadata.tsv
dataset-referred-to-by-all-metadata-tsvs/
dataset-only-referred-to-by-histology/
```

...something like this would not be:

```
visium-metadata.tsv
histology-metadata.tsv
rnaseq-metadata.tsv
codex-metadata.tsv
dataset-referred-to-by-all-metadata-tsvs-except-codex/
dataset-only-referred-to-by-histology/
dataset-only-referred-to-by-codex/
```

In the above case, you would get the error `Invalid child assay type(s) for parent type Visium (no probes): ["CODEX"].`.

I think that if you can submit multiple dataset types in a given upload then the multi-assay case above should be fixed, but that's how it works right now.

sunset666:
> > Since this is just going to work on Uploads, I'm expecting it won't affect the reorganization process (because it will mainly create the datasets specified in the big files, which all will be pointing to the primary dataset folder on that upload). And the split of components don't really move anything, they just get create the 'virtual' dataset and the data path will point to the primary.
> 
> But are we detecting that the second Histology dataset is not technically part of a Multiome dataset? In Gesina's example, we would have two lines in the Histology TSV, and one line each in Visium and RNAseq TSVs. That second line in the Histology TSV which points to `dataset-2` should be treated as a non-multiome dataset. I don't think that this scenario is currently supported by the metadata extraction classes.

We are not even using the components metadata as part of the reorganization, so, in case there is a dataset referenced by any of those that is not part of a "full" dataset, that will remain in the upload directory. We are just using the component-metadata to build the individual metadata for that particular component. If we are seeing independent datasets added to a Multi upload, that will indeed be out of the current scope.

jpuerto-psc:
> We are not even using the components metadata as part of the reorganization, so, in case there is a dataset referenced by any of those that is not part of a "full" dataset, that will remain in the upload directory. We are just using the component-metadata to build the individual metadata for that particular component. If we are seeing independent datasets added to a Multi upload, that will indeed be out of the current scope.

Understood. When others are back from break, lets discuss whether it is okay for us to be more strict with our uploads such that we do **not** allow these situations to be valid. This will require some changes to the validation @gesinaphillips if this is  indeed how we want to proceed. Have to confirm with Phil/Stephen/Brendan.

jpuerto-psc:
As per conversation on 01/05 - we will **not** support mixed uploads (IE - 10x and single RNAseq datasets). Tasks added for detecting these situations during validation.

src/ingest-pipeline/misc/tools/survey.py
jpuerto-psc: As per 01/05, we can merge these changes into devel so that Sunset can test changes on DEV. We will continue this discussion on 01/11, but that should not block Sunset's testing.

src/ingest-pipeline/misc/tools/split_and_create.py
jpuerto-psc: As per conversation on 01/05 - we will not support mixed uploads (IE - 10x and single RNAseq datasets). Tasks added for detecting these situations during validation.

jpuerto-psc (approved):

src/ingest-pipeline/misc/tools/survey.py
jpuerto-psc: Always send data_types and dataset_type, but send an empty string for the data_types field for new CEDAR style datasets.

we need a mapping of old non-CEDAR style dataset information to new CEDAR dataset_type. 

Ex:
- ATACseq-bulk -> ATACseq

PR #813, opened 2024-01-10
Nihdev 465 update rebuild metadata dag for soft assays

jpuerto-psc:
Two changes here:

- Change how the assaytype invocation works such that the request to the assay classifier and the assertion are not coupled
- Create a new DAG that takes a list of uuids that can be either primary or processed and spawns the appropriate DAG for that primary or processed dataset.


jpuerto-psc:
@sunset666 - let me know what you think about moving the rebuild_metadata in this direction. I have no clue whether this will work or not but would like to test it on devel if you feel that this is the path we want to take moving forward.

PR #821, opened 2024-01-17
Nihdev 465 update rebuild metadata dag for soft assays

jpuerto-psc:
This PR adds functionality to process a bulk set of UUIDs (independent of whether they are derived or primary). 

There is one DAG (`rebuild_multiple_metadata`) which accepts a list of UUIDs. This DAG parses that list of UUIDs and splits it into a list of primary datasets and a list of processed datasets. It then uses the new Airflow dynamic tasking functionality to spawn a DAG for each dataset in the list, based on whether it is primary or processed. The `rebuild_primary_dataset_metadata` DAG is a direct clone of the pre-existing `rebuild_metadata` DAG. The `rebuild_processed_dataset` is just an EmptyOperator for now as we don't have a clear idea of how we should get the appropriate metadata into the dataset. However, this allows us to prove that datasets are going into the appropriate 'rebuilding' DAG.

Graph when a mix of primary and processed UUIDs are provided:
image

Graph when only primary UUIDs are provided:
image

Graph when only processed UUIDs are provided:
image

A few caveats, there are some changes to the way that I grab soft assay data that will have to be incorporated into the branch for reorganizing multi-assay uploads. Also, the `rebuild_multiple_metadata` DAG does **not** use the HMDAG class, as this was causing issues when spawning the dynamic tasks. This should not cause any issues as far as I am aware.


src/ingest-pipeline/airflow/dags/rebuild_multiple_metadata.py
```diff
@@ -0,0 +1,115 @@
...
+
+            # If we got nothing back from soft_data, then let's try to determine using entity_api
+            if soft_data:
+                if soft_data.get('primary'):
+                    kwargs['dag_run'].conf['primary_datasets'].append(uuid)
+                else:
+                    kwargs['dag_run'].conf['processed_datasets'].append(uuid)
+            else:
+                my_callable = lambda **kwargs: uuid
+                ds_rslt = pythonop_get_dataset_state(dataset_uuid_callable=my_callable, **kwargs)
```
gesinaphillips, line 56: `# Nitpicky`
Could just pass the lambda as the arg to dataset_uuid_callable since it's only used the once?
`ds_rslt = pythonop_get_dataset_state(dataset_uuid_callable=lambda **kwargs: uuid, **kwargs)`

src/ingest-pipeline/airflow/dags/rebuild_multiple_metadata.py
```diff
@@ -0,0 +1,115 @@
...
+    def build_dataset_lists(**kwargs):
+        kwargs['dag_run'].conf['primary_datasets'] = []
+        kwargs['dag_run'].conf['processed_datasets'] = []
+
+        print('dag_run conf follows:')
+        pprint(kwargs['dag_run'].conf)
+        for uuid in kwargs['dag_run'].conf['uuids']:
+            soft_data = get_soft_data(uuid, **kwargs)
+
+            # If we got nothing back from soft_data, then let's try to determine using entity_api
```
gesinaphillips, line 48: Is it meaningful if we don't get something back from the soft assay endpoint? Do we want to log this somewhere?

gesinaphillips (approved):

src/ingest-pipeline/airflow/dags/rebuild_multiple_metadata.py
jpuerto-psc: We can log this, although I can see this being a common scenario where the dataset that we're rebuilding metadata for has no metadata (so it doesn't return any matches)

sunset666 (approved):

PR #841, opened 2024-02-06
Jpuerto/example8 dag

jpuerto-psc:
This PR adds a DAG to generate @icaoberg daily report in py-hubmapbags

Deployment of this PR will require that we update our conda environments @sunset666


sunset666 (commented):
@jpuerto-psc, I don't see any updates to the requirements.txt. Not sure why the conda environments needs to be updated then.

jpuerto-psc:
@sunset666 - the py-hubmapbags submodule had its requirements.txt updated

sunset666:
@jpuerto-psc, if that's the case we need to add those changes to the main requirements.txt, we haven't found a way to recursively install requirements from submodules, yet.

PR #858, opened 2024-03-27
DAGs: Change SnapATAC to ArchR for the pipeline shorthand.

jpuerto-psc:
This is to make sure that we are using the latest pipeline description moving forward.


sunset666 (approved):

jpuerto-psc:
@mruffalo confirmed that new pipeline runs moving forward should use this name.

PR #870, opened 2024-05-07
Jpuerto/dc helper dag

jpuerto-psc:
This PR adds a helper DAG for @bhonick to be able to bulk update status, ingest_task, assigned_to_group_name, and description for uploads and datasets.


src/ingest-pipeline/airflow/dags/bulk_update_entities.py
```diff
@@ -0,0 +1,127 @@
...
+        for uuid in uuids:
+            # If this fails out then we know its something other than an upload or dataset
+            try:
+                pythonop_get_dataset_state(dataset_uuid_callable=lambda **kwargs: uuid, **kwargs)
+                filtered_uuids.append(uuid)
+            except Exception as e:
+                print(f"{uuid} is neither a dataset nor an upload and will be skipped.")
+                print(repr(e))
+
+        kwargs["ti"].xcom_push(key="uuids", value=filtered_uuids)
```
sunset666, line 79: There is probably no point in storing the uuids in the Airflow DB if they are not going to be used downstream.

src/ingest-pipeline/airflow/dags/bulk_update_entities.py
```diff
@@ -0,0 +1,127 @@
...
+
+        for uuid in uuids:
+            endpoint = f"entities/{uuid}"
+            try:
+                response = http_hook.run(endpoint, headers=headers, data=json.dumps(metadata))
+                print("response: ")
+                pprint(response.json())
+                time.sleep(10)
+            except:
+                print(f"ERROR: UUID {uuid} could not be updated.")
```
sunset666, line 114: Might as well have a sleep in the exception?

sunset666 (changes requested):

src/ingest-pipeline/airflow/dags/bulk_update_entities.py
jpuerto-psc: Comments addressed

src/ingest-pipeline/airflow/dags/bulk_update_entities.py
jpuerto-psc: Moved sleep to outside the try/except block

sunset666 (approved):

PR #878, opened 2024-05-28
S&C: Fix the non_global_files issues

jpuerto-psc:


sunset666 (changes requested):
I just tested this and it will throw an exception:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'numpy.float64' object has no attribute 'isnull'
```

sunset666 (approved):

PR #888, opened 2024-06-27
S&C: Add check to make sure that the destination path does not yet ex…

jpuerto-psc:
…ist.


src/ingest-pipeline/misc/tools/split_and_create.py
```diff
@@ -339,16 +339,16 @@ def copy_contrib_antibodies(dest_extras_path, source_entity, old_paths, dryrun):
             print(f"copy {old_path} to {dest_extras_path}")
         else:
             src_path = source_entity.full_path / old_path
+            dest_path = dest_extras_path / old_path.name
 
-            if src_path.exists():
+            if src_path.exists() and not dest_path.exists():
                 dest_extras_path.mkdir(parents=True, exist_ok=True)
                 copy2(src_path, dest_extras_path / old_path.name)
```
gesinaphillips, line 346: Can replace second arg to copy2 with `dest_path`

gesinaphillips (approved):

PR #898, opened 2024-07-11
DAG to find and change file extensions, defaults to TIF -> TIFF

gesinaphillips:
By default, running this DAG with one or more UUIDs in the `uuid_list` of the config will find .tif/.TIF files and convert them to .tiff.

Written more abstractly to accommodate a) specifying a different target/replacement pair, b) just find and log file paths without replacing.


src/ingest-pipeline/airflow/dags/standardize_extensions.py
```diff
@@ -0,0 +1,183 @@
...
+        if not ds_rslt:
+            raise AirflowException(f"Invalid uuid/doi for group: {uuid}")
+
+        for key in ["entity_type", "status", "uuid", "local_directory_full_path"]:
+            assert key in ds_rslt, f"Dataset status for {uuid} has no {key}"
+
+        if not ds_rslt["entity_type"] in ["Upload"]:
+            raise AirflowException(f"Entity {uuid} is not an Upload")
+
+        if not ds_rslt["status"] in ["New", "Invalid", "Error"]:
```
jpuerto-psc, line 74: This is fine for now, but we may want to extend the status set based on feedback from Brendan.

jpuerto-psc (approved):

src/ingest-pipeline/airflow/dags/standardize_extensions.py
gesinaphillips: Oh right, yes, I just guessed here, I meant to check! Happy to update in future.

PR #975, opened 2024-11-08
Add 3D IMC segmentation

mruffalo:


src/ingest-pipeline/airflow/dags/imc_3d_deepcell_3dcc.py
```diff
@@ -0,0 +1,531 @@
...
+        print("data_dir: ", data_dir)
+
+        workflow = cwl_workflows["segmentation"]
+        meta_yml_path = workflow.parent / "meta.yaml"
+
+        command = [
+            *get_cwltool_base_cmd(tmpdir),
+            # '--singularity',
+            workflow,
+            "--gpus=all",
```
sunset666, line 90: the pipeline.cwl doesn't have gpus as a parameter.

src/ingest-pipeline/airflow/dags/imc_3d_deepcell_3dcc.py
```diff
@@ -0,0 +1,531 @@
...
+            "--cytoplasm_markers",
+            "In115",
+            "--membrane_markers",
+            "La139" "--nucleus_markers",
+            "Ir191" "--input_dir",
```
sunset666, line 95: Probably missing ',' after La139 and Ir191?

sunset666 (changes requested):

src/ingest-pipeline/airflow/dags/imc_3d_deepcell_3dcc.py
mruffalo: Fixed; thanks!

src/ingest-pipeline/airflow/dags/imc_3d_deepcell_3dcc.py
mruffalo: Correct. GPU requirements apply to each individual `CommandLineTool` in a CWL workflow, not a `Workflow` that composes them together. The segmentation step has the GPU requirement: https://github.com/hubmapconsortium/imc-3d-segmentation/blob/main/steps/segmentation.cwl#L7

sunset666: yes, but it need to be declared so it can be passed along, currently it throws an error because of that.
usage: /hive/users/hive/hubmap/vm001-dev/ingest-pipeline/src/ingest-pipeline/airflow/dags/cwl/imc-3d-segmentation/pipeline.cwl
       [-h] --cytoplasm_markers CYTOPLASM_MARKERS --input_dir INPUT_DIR
       --membrane_markers MEMBRANE_MARKERS --nucleus_markers NUCLEUS_MARKERS
       [--processes PROCESSES]
       [job_order]
.../imc-3d-segmentation/pipeline.cwl: error: unrecognized arguments: --gpus=all

https://github.com/hubmapconsortium/mibi-pipeline/blob/7501938b53db117551ff02ca5597256d6795852e/pipeline.cwl#L8

mruffalo: Ah, that was my mistake in making a copy of the MIBI DAG and then modifying appropriately -- I'd forgotten that this workflow doesn't accept a `--gpus` parameter at all. I'll remove it from the DAG as soon as I can.

PR #1041, opened 2025-03-13
New flow: creating datasets just after sucessful runs

sunset666:
* A new processed dataset will only be created after a successful pipeline run.
* Support for containers to be created on the fly.
* Bump CODEX v2.7.1.
* New CODEX cell types DeepCellTypes and RIBCA.


jpuerto-psc (approved):

gesinaphillips (approved):
Very small comment about updating example config, if applicable.

PR #1045, opened 2025-03-31
Bugfix metadata update on components

sunset666:
Updating utils and StatusChanger to avoid updating status on components.
Small bugfix on StatusChange to better handle of no status updates.


gesinaphillips (approved):
This seems like a good change, thank you!

jpuerto-psc (approved):

PR #1057, opened 2025-05-22
Sfd5311/alignment json fix

SFD5311:


jpuerto-psc:
@SFD5311 - just to confirm, this should cancel out #1053 #1051 and #1038 ?

PR #1085, opened 2025-07-22
Welling/zarr zipper

jswelling:
Adding a tool to recursively find and zip up zarr directories

### DIFF
--- a/src/ingest-pipeline/misc/tools/zarr_zipper.py
+++ b/src/ingest-pipeline/misc/tools/zarr_zipper.py
@@ -1,0 +1,94 @@
+#! /usr/bin/env python
+
+import argparse
+import re
+from pathlib import Path
+from pprint import pprint
+from typing import List, Tuple, TypeVar, Union
+import logging
+
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+UUID_REGEX = re.compile(r"[0123456789aAbBcCdDeEfF]{32}")
+
+def parse_path(path) -> Tuple[Path, str, str]:
+    LOGGER.debug(f"starting with {path}")
+    target_path = path.parent
+    uuid_path = target_path
+    elts = list(uuid_path.parts)
+    LOGGER.debug(f"elts: {elts}")
+    elts.reverse()
+    for elt in elts:
+        LOGGER.debug(f"testing {elt}")
+        if UUID_REGEX.match(str(elt)):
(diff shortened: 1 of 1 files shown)


src/ingest-pipeline/misc/tools/zarr_zipper.py
```diff
@@ -0,0 +1,130 @@
...
+        help="describe the steps that would be taken but" " do not make changes",
+        action="store_true",
+    )
+
+    args = parser.parse_args()
+
+    dryrun = args.dryrun
+
+    start_path = Path(args.starting_path)
+    if not start_path.is_dir():
```
jpuerto-psc, line 118: Can we add a check to make sure its not operating from the root directory as a starting point

src/ingest-pipeline/misc/tools/zarr_zipper.py
```diff
@@ -0,0 +1,130 @@
...
+    zip -r ../reg001_expr-anndata.zarr.zip .
+    cd ..
+    mv reg001_expr-anndata.zarr ../../672111342c0002f90e1cb1bc5686ecf1_anndata-zarr_reg001_expr-anndata.zarr
+    """
+    full_target_path = target_path / zarr_name
+    zip_name = target_path / f"{zarr_name}.zip"
+    cmd1 = f"cd '{full_target_path}'"
+    cmd2 = f"zip -r '{zip_name}' ."
+    cmd3 = "cd .."
+    cmd4 = f"rm -r -v {zarr_name}"
```
jpuerto-psc, line 51: Can we add a flag for the removal? IE, only remove if we set that flag specifically.

src/ingest-pipeline/misc/tools/zarr_zipper.py
```diff
@@ -0,0 +1,130 @@
...
+    if not start_path.is_dir():
+        parser.error(f"{start_path} is not an existing directory")
+
+    all_valid_uuids = []
+    for candidate in start_path.glob(args.proto):
+        uuid = restructure(candidate, dryrun=dryrun)
+        if verify_uuid(uuid):
+            all_valid_uuids.append(uuid)
+
+    print(f"rebuilding json: {json.dumps({'uuids':all_valid_uuids})}")
```
jpuerto-psc, line 127: Is this what will be passed to the metadata rebuilder DAG?

shirey (approved):

PR #1108, opened 2025-09-10
fix: new schema file for rebuild metadata

jpuerto-psc:
This PR adds a new YAML schema and updates DAGs to use this new YAML schema.

### DIFF
--- a/src/ingest-pipeline/schemata/rebuild_metadata_schema.yml
+++ b/src/ingest-pipeline/schemata/rebuild_metadata_schema.yml
@@ -1,0 +1,17 @@
+'$schema': 'http://json-schema.org/schema#'
+'$id': 'http://schemata.hubmapconsortium.org/launch_checksums_schema.yml'
+'title': 'launch_checksums metadata schema'
+'description': 'launch_checksums metadata schema'
+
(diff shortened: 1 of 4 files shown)


src/ingest-pipeline/schemata/rebuild_metadata_schema.yml
```diff
@@ -0,0 +1,17 @@
+'$schema': 'http://json-schema.org/schema#'
+'$id': 'http://schemata.hubmapconsortium.org/launch_checksums_schema.yml'
```
jswelling, line 2: These launch_checksums strings should get changed as well?

src/ingest-pipeline/schemata/rebuild_metadata_schema.yml
```diff
@@ -0,0 +1,17 @@
...
+
+'definitions':
+
+  'launch_checksums':
+     'type': 'object'
+     'properties':
+        'uuid':
+          'type': 'string'
+          'description': 'a dataset uuid'
+     'required': ['uuid']
```
jswelling, line 16: If lz_path and src_path are required, should they be here as well?

jswelling (changes requested):
At least the schema name strings should probably be cleaned up.

PR #1253, opened 2026-04-16
Refactored pythonop_send_create_dataset

gesinaphillips:
tl;dr: Got a non-JSON response from Ingest API in a case where the metadata was too large. The goal of this work was to:
- Fix the issue
- Catch JSONDecodeErrors

I did that and some refactoring. (Type-checker users, rejoice.)

----

Ultimately, the issue was that `pythonop_send_create_dataset` was calling ingest-api to retrieve the parent dataset info. Although `exclude=direct_ancestors.files` was included in that call, that param was not passed on to ingest-api's call to entity-api. This caused the response to be too large for ingest-api to return, so it returned a message in the `text` attribute. I solved this by calling entity-api directly with the appropriate param.

### Fixed case where `pythonop_send_create_dataset` fails due to large response.
- Refactoring function during troubleshooting to remove overlong try/except block.
- Abstracted request code into a reusable function (`make_httphook_request`).
     - Added error logging for non-JSON responses.
     - Added logging truncation to `make_httphook_request` to prevent laggy Airflow logs.
- Integration of helper function `get_submission_context` should ensure this issue does not happen again.

### Surface-level refactor of utils.py.
- Replaced repeated request code in other functions with `make_httphook_request` / `get_submission_context`.
- Added `get_env` function to consolidate environment checking.
- Modernized typing (removed List, Dict, Union, Optional).
- Finally cleaned up the typing issues identified by my type checker (e.g. issues with functions using generic type `StrOrListStr`, casting data to the correct type before performing other operations, handling null cases, etc.).

### Other
- Added Publication QA to "good statuses" for email messaging.

### DIFF
--- a/src/ingest-pipeline/airflow/dags/utils.py
+++ b/src/ingest-pipeline/airflow/dags/utils.py
@@ -27,7 +27,6 @@ from typing import (
 )
 
 import cwltool  # used to find its path
-import requests
 import yaml
 from cryptography.fernet import Fernet
 from requests import codes
@@ -870,8 +869,6 @@ def make_httphook_request(
     method: str = "GET",
     payload: dict | None = None,
     headers: dict | None = None,
-    return_text_response: bool = False,
-    log: bool = True,
 ) -> dict:
     assert token or headers, "Must pass in either a token or header dict"
     if not headers:
@@ -894,18 +891,19 @@ def make_httphook_request(
         raise RuntimeError(f"misc error {e} on {endpoint}")
     try:
         response_json = response.json()
-        if log:
-            print(f"Response from '{method}' on {endpoint} ({http_conn_id}):")
+        print(f"Response from '{method}' on {endpoint} ({http_conn_id}):")
+        if content_len := response.get("Content-Length"):
+            if int(content_len) > 500000:  # length chosen arbitrarily, feel free to change
+                print(f"Response too long, not logging.")
+        else:
             print(response_json)
         return response_json
     except JSONDecodeError as e:
-        # Will raise unless response has `text` property and `return_text_response=True` (non-default)
+        # Will raise unless response has `text` property
         if response.text:
             print(
                 f"Received non-JSON response. Text: {response.text}; content: {response.content}"
             )
-            if return_text_response:
-                return response.text
         raise
 
 
@@ -1142,7 +1140,7 @@ def pythonop_get_dataset_state(**kwargs) -> dict:
     auth_tok = get_auth_tok(**kwargs)
 
     try:
-        ds_rslt = get_submission_context(auth_tok, uuid, log=False)
+        ds_rslt = get_submission_context(auth_tok, uuid)
     except Exception:
         return {}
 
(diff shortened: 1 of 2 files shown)


src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -2060,52 +1965,60 @@ def gather_calculated_metadata(**kwargs):
...
         )
-        response.raise_for_status()
-        return response.json()
-    except HTTPError as e:
-        print(f"ERROR: {e}")
-        if e.response.status_code == codes.unauthorized:
-            raise RuntimeError("entity database authorization was rejected?")
-        print("benign error")
-        return {}
+    except JSONDecodeError:
```
gesinaphillips: Actual fix for dealing with text response when metadata is too large.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -874,6 +862,51 @@ def get_auth_tok(**kwargs) -> str:
     return auth_tok
 
 
+def make_httphook_request(
```
gesinaphillips, line 865: New func.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -874,6 +862,50 @@ def get_auth_tok(**kwargs) -> str:
...
+        return response_json
+    except JSONDecodeError as e:
+        if response.text:
+            print(
+                f"Received non-JSON response. Text: {response.text}; content: {response.content}"
+            )
+        raise
+
+
 def pythonop_send_create_dataset(**kwargs) -> str:
```
gesinaphillips, line 909: Major refactor.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -1937,7 +1859,7 @@ def get_preserve_scratch_resource(dag_id: str) -> bool:
     return bool(rec["preserve_scratch"])
 
 
-def get_threads_resource(dag_id: str, task_id: Optional[str] = None) -> int:
+def get_threads_resource(dag_id: str, task_id: str | None = None) -> int:
```
gesinaphillips, line 1862: Broke up logic in get_threads_resource because uncaught null cases were making my type checker go bonkers.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -2060,103 +1965,42 @@ def gather_calculated_metadata(**kwargs):
...
-    response.raise_for_status()
+    payload = {"message": message, "channel": channel}
+    return make_httphook_request(
+        "notify", "ingest_api_connection", token, method="POST", payload=payload
+    )
 
 
-def env_appropriate_slack_channel(prod_channel: str) -> str:
-    default = "C0A8ES4M9RU"  # test-notifications
+def get_env() -> str:
```
gesinaphillips, line 1974: New func.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -2239,5 +2079,56 @@ def send_email(
     return True
 
 
+def main():
```
gesinaphillips, line 2082: Code untouched, just moved to be below all utils funcs.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -2060,103 +1965,42 @@ def gather_calculated_metadata(**kwargs):
...
+
+def env_appropriate_slack_channel(prod_channel: str) -> str:
+    if get_env() == "prod":
         return prod_channel
-    return default
+    return "C0A8ES4M9RU"  # test-notifications
 
 
-# This is simplified from pythonop_get_dataset_state in utils
 def get_submission_context(token: str, uuid: str, headers: dict | None = None) -> dict[str, Any]:
```
gesinaphillips, line 1993: More helpful helper func.

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -210,6 +205,7 @@ class DummyFileMatcher(FileMatcher):
     """
 
     def get_file_metadata(self, file_path: Path) -> ManifestMatch:
+        del file_path
```
jpuerto-psc, line 208: what is this `del` doing?

src/ingest-pipeline/airflow/dags/utils.py
```diff
@@ -308,7 +304,7 @@ def get_cwl_cmd_from_workflows(
     if not outdir_present:
         command.extend(["--outdir", str(tmp_dir / "cwl_out")])
 
-    command.append(Path(workflow["workflow_path"]))
+    command.append(workflow["workflow_path"])
```
jpuerto-psc: are we sure that removing the Path casting won't cause issues?

src/ingest-pipeline/airflow/dags/utils.py
gesinaphillips: Perhaps slightly controversial decision (or maybe there's a better way to handle it that I don't know of): it's a required parameter for FileMatcher, but is unused here, so deleting quiets any overzealous unused arg checking.

src/ingest-pipeline/airflow/dags/utils.py
gesinaphillips: You know what, let's be safe and put it back.

PR #1257, opened 2026-04-17
Jmuerto/multithread scrubbing

jpuerto-psc:


gesinaphillips (approved):

gesinaphillips (commented):
Tests failing because `resource_map.yml` is missing `preserve_scratch` property.

src/ingest-pipeline/misc/tools/scrub_fastqs.py
```diff
@@ -35,7 +34,7 @@ def _run_scrubber(input_path: Path, output_name: str) -> None:
     )
 
 
-def _scrub_fastq(fastq_path: Path) -> None:
+def _scrub_fastq(fastq_path: Path, num_threads: int = 1) -> None:
```
sunset666, line 37: Is it for future development? no use for num_threads in the function.

jpuerto-psc: I just didn't remove it, I was going to pass it down to the docker invocation but the docs indicated that the docker run tries to use as many cores as possible, so no need to pass anything down there.

PR #1283, opened 2026-05-20
Removing unused imports Bugfix Phenocycler Cell Sized final steps flow

sunset666:


src/ingest-pipeline/airflow/dags/reorganize_upload.py
```diff
@@ -37,15 +37,15 @@
     get_queue_resource,
     get_preserve_scratch_resource,
     get_soft_data_assaytype,
-    get_threads_resource,
+    # get_threads_resource,
```
gesinaphillips, line 40: Just curious about comments vs. removals here and in celldive_deepcell!

gesinaphillips (approved):

src/ingest-pipeline/airflow/dags/reorganize_upload.py
sunset666: Since there is still the option to use the Stellar pre-convert for some data (commented code), the imports were commented as well.

PR #1309, opened 2026-06-23
Welling/ro crates

jswelling:
This PR provides infrastructure to capture the --provenance output of cwltool in an ro-crate object.  Future work will be needed to integrate existing file descriptions, etc. into the crate, to make sure the crate contains all the relevant info, and hopefully to reduce the volume of the crate.  The --provenance flag doesn't appear to work on the first cwltool call in the chain.  This will also need to be investigated.
A new class, CrateManager, tracks the information necessary to build the crate.  Mods to utils and to the dag itself pull in the functionality of this new class.  An example dag, ometiff_pyramid_ims_crate.py , is provided as an example of how to modify a DAG to produce and ro-crate.


src/ingest-pipeline/airflow/dags/crate_manager.py
```diff
@@ -0,0 +1,148 @@
...
+            logging.debug(f"updated from previous incarnation {prev_info}")
+        else:
+            logging.debug("no previous incarnation found")
+        crate = ROCrate()
+        crate.name = "Aggregated Collection of BagIt Datasets"
+        crate.description = "An overarching RO-Crate grouping multiple domain-specific BagIt objects."
+
+        for path in [Path(p) for p in self.crate_chain]:
+            if path.is_dir():
+                is_bag = False
```
gesinaphillips, line 102: I don't think you need `is_bag` because in the try/except you either set `bag` or continue to the next path in the loop (or raise if it's not one of the specified exception types). So I don't think you should be able to get to line 111 without `bag` being set.

gesinaphillips (approved):

PR #1335, opened 2026-07-29
Error status resetting

gesinaphillips:
- Allows `make_send_status_msg_function` to reset Error status
- Allows `StatusChanger` to reset primary dataset status from Error->QA when a derived dataset reaches QA (or New, but that probably won't hit the `StatusChanger` I don't think)
     - Are we ever creating multiple derived datasets from the same primary simultaneously? This might hide failures if so.
- I believe we tabled the issue we were worried about where rerunning the `send_status_msg` DAG step might create duplicate provenance metadata. Instead of rerunning `send_status_msg`, reset the status manually if a DAG fails at or after the `send_status_msg` task (and restart subsequent tasks).

### DIFF
--- a/src/ingest-pipeline/airflow/dags/status_change/status_manager.py
+++ b/src/ingest-pipeline/airflow/dags/status_change/status_manager.py
@@ -367,9 +367,12 @@ class StatusChanger(EntityUpdater):
         if primary_status != "error":
             return
         try:
-            logging.info(f"Updating primary dataset status from {primary_status} to QA.")
+            logging.info(f"Updating primary dataset status from Error to QA.")
             response = put_request_to_entity_api(
-                primary_uuid, self.token, {"status": "QA"}, {"reindex-priority": self.reindex}
+                primary_uuid,
+                self.token,
+                {"status": "QA", "pipeline_message": ""},
+                {"reindex-priority": self.reindex},
             )
             logging.info(f"Response: {response}")
         except Exception as e:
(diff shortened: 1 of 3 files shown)


src/ingest-pipeline/airflow/dags/status_change/status_manager.py
```diff
@@ -342,6 +351,32 @@ def _send_to_entity_updater(self, log_msg: str):
...
+                f"Can't find primary dataset for derived dataset {self.uuid}, not updating."
+            )
+        primary_metadata = get_submission_context(self.token, primary_uuid)
+        primary_status = primary_metadata.get("status", "").lower()
+        if primary_status != "error":
+            return
+        try:
+            logging.info(f"Updating primary dataset status from {primary_status} to QA.")
+            response = put_request_to_entity_api(
+                primary_uuid, self.token, {"status": "QA"}, {"reindex-priority": self.reindex}
```
sunset666: Maybe adding pipeline_message: "", to clean any error message it got?

sunset666 (changes requested):

src/ingest-pipeline/airflow/dags/status_change/status_manager.py
gesinaphillips: Oh duh, thank you!!

jpuerto-psc (approved):

src/ingest-pipeline/airflow/dags/status_change/status_manager.py
gesinaphillips: Updated!

sunset666 (approved):

